PR #4, opened 2022-08-15
Add webpack for build and add inital SAM template

andylangridge:
Added a webpack config for the "build", and a SAM template for the existing hello world lambda function we have.

Webpack will transpile the Typescript code into Javascript in the `dist` directory. It will also include source maps. In order to make the source maps work, the Lambda function needs the `--enable-source-maps` flag for Node options. Have set this as a global environment variable so all Lambdas should inherit it.

Additionally, some things to point out in `package.json`:
* Have removed the `"type": "module"` line - I think this is ok as we can use all the ECMAScript features by specifying in tsconfig (which is the case).
* Removed `aws-lambda` as a dependency - Think we don't need this as this package is for deploying lambdas, and we are using SAM.


rikumolugds (commented):
Just wondering whether there were more files we could add to .gitignore.

andylangridge:
> 

If you're referring to the files in the `.yarn` folder, these are there on purpose. Yarn compresses its dependencies, so they can be committed. The idea is that they are ready to go out the box, making things much faster.

gdsandrewmccafferty:
When writing lambda handlers, I often have things like this:

import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';

Doesn't that mean we should keep aws-lambda as a dependency, just so we can import the types for requests etc?

andylangridge:
> When writing lambda handlers, I often have things like this:
> 
> import { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
> 
> Doesn't that mean we should keep aws-lambda as a dependency, just so we can import the types for requests etc?

These are available via @types/aws-lambda :)

SamBerry-GDS (approved):

PR #8, opened 2022-08-18
Add an assume role script

andylangridge:
Adds a script we can use with the AWS CLI to assume a role in another AWS account.

### DIFF
--- a/scripts/assumeRole.sh
+++ b/scripts/assumeRole.sh
@@ -11,7 +11,12 @@
 # source ./scripts/assumeRole.sh 123456
 MFA_CODE=$1
 
-# Script assumes you are authenticated with the 'gds-users' AWS account via the AWS CLI
+# Script assumes you are authenticated with the 'gds-users' AWS account via the AWS credentials file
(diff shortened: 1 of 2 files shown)


scripts/assumeRole.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# To run this script requires you to have the following environment variables set:
+# GDS_EMAIL_ADDRESS - the email address of the account for the 'gds-users' AWS account
+# ROLE_ARN - the ARN of the role you want to assume in the relevant AWS account
+: "${GDS_EMAIL_ADDRESS:?GDS_EMAIL_ADDRESS not set or empty}"
```
gdsandrewmccafferty, line 6: When trying it out, I noticed that the error message includes the name of the variable anyway
`./assumeRole.sh:6: GDS_EMAIL_ADDRESS: GDS_EMAIL_ADDRESS not set or empty`

, handily, so I think you can leave it out of the message after the `?`

andylangridge: Fixed!

gdsandrewmccafferty (approved):

PR #10, opened 2022-08-19
Add deployment pipeline

andylangridge:
- Added a pipeline that can deploy to AWS
- Deployment job re-uses the build artifact from the validation job
- Includes some secrets that have been added to this repos config
- Added SAM template to Prettier ignore file as when conditions are added it looks like it was breaking the layout
- Updated SAM template with Code Signing config - the values for this are provided by the Codebuild pipeline that gets triggered when this Deploy job uploads the SAM package to AWS. We don't so `sam deploy` in GH Actions, instead the final is a `sam package` that uploads to S3, Secure Pipelines handles the rest.

### DIFF
--- a/.github/workflows/pipeline.yaml
+++ b/.github/workflows/pipeline.yaml
@@ -5,7 +5,7 @@ on:
     branches: [main]
   push:
     branches: [main]
-  workflow_dispatch:
+  workflow_dispatch: {}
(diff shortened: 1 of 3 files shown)


gdsandrewmccafferty (approved):
LGTM. 🦅

PR #11, opened 2022-08-22
Use exact version dev dependencies

SamBerry-GDS:
Removes approximate version ranges for package dependencies inline with DI architecture dependency guidance: https://github.com/alphagov/digital-identity-architecture/blob/main/adr/0037-nodejs-npm-production-dependencies.md


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
andylangridge, line 1: This file is causing the PnP files to be deleted, which will stop us from the zero-installs feature.

SamBerry-GDS: Reverted

andylangridge (approved):

PR #13, opened 2022-08-22
Added Zendesk webhook

gdsandrewmccafferty:


template.yaml
```diff
@@ -28,7 +28,13 @@ Globals:
     Runtime: nodejs16.x
 
 Resources:
-  HelloWorld:
+  ZendeskWebhook:
```
andylangridge, line 31: Could we call this `ValidateRequestFunction` or something along those lines? Assuming that this is the first function in our flow, and it's going to validate the details in the Zendesk ticket.

template.yaml
```diff
@@ -28,7 +28,13 @@ Globals:
...
     Type: AWS::Serverless::Function
     Properties:
-      Handler: helloWorld.handler
+      Handler: zendeskWebhook.handler
```
andylangridge, line 34: Same as the comment above, could this just be `validateRequest.handler` if we think that's a good name.

src/testUtils/events/defaultApiRequest.ts
```diff
@@ -0,0 +1,49 @@
+export const defaultApiRequest = {
```
andylangridge, line 1: You don't need the entire object for the event I think. Just the stuff that's going to be useful will do!

andylangridge: Also would this be better as just a standard JSON file? I think the convention is to stick it in an `events` folder at the root. That way you can use it with `sam local invoke`.

andylangridge (approved):

PR #23, opened 2022-08-23
Wrote pseudocode outlining next steps in data request

mabon-gds:


src/handlers/initiateDataRequest.ts
```diff
@@ -0,0 +1,29 @@
...
+    // inform Zendesk
+    updateZendeskTicket(event.body, 'invalid params')
+    return {
+      statusCode: 400,
+      body: JSON.stringify({
+        message: 'Request parameters invalid'
+      })
+    }
+  }
+  initiateS3DataCheck(validatedZendeskRequest.dataRequestParams)
```
andylangridge, line 22: Just for my own understanding really, but why would you not make this function asynchronous? I often see that functions are async in Lambda code, so wondering if there was any particular reason you wouldn't do that here?

gdsandrewmccafferty: Yep, good point! Both those things will have some network access

andylangridge (dismissed):

gdsandrewmccafferty (approved):

PR #30, opened 2022-08-30
List objects in both audit and analysis buckets

andylangridge:
* Added the Analysis bucket as a resource to the SAM template
* Check what data is in both S3 buckets, and calculate what objects to copy
* Moved the testUtils folder into a new utils folder (where I added a file for constants, and a file for helper functions)


src/services/listS3Objects.ts
```diff
@@ -0,0 +1,26 @@
...
+  S3Client,
+  ListObjectsV2Command,
+  ListObjectsV2CommandInput
+} from '@aws-sdk/client-s3'
+
+export const listS3Objects = async (
+  input: ListObjectsV2CommandInput,
+  objects: string[] = []
+): Promise<string[]> => {
+  const client = new S3Client({ region: 'eu-west-2' })
```
SamBerry-GDS, line 11: Would it make sense for this to be an env var

src/services/listS3Objects.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  test('response has no continuation token', async () => {
+    s3Mock.on(ListObjectsV2Command).resolves({
+      Contents: [{ Key: 'example-object' }]
+    })
+
+    const result = await listS3Objects(input)
+    expect(result).toEqual(['example-object'])
+  })
+
+  test('has continuation token', async () => {
```
SamBerry-GDS, line 29: Could we make these slightly more descriptive e.g. Calls with continuation tokens return results for all pages

src/services/listS3Objects.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  test('respone has no next continuation token', async () => {
+    s3Mock.on(ListObjectsV2Command).resolves({
+      Contents: [{ Key: 'example-object' }],
+      ContinuationToken: 'page2'
+    })
+
+    const result = await listS3Objects(input)
+    expect(result).toEqual(['example-object'])
+  })
```
SamBerry-GDS, line 62: Is it possible to get no continuation or content and if is it worth testing?

src/services/listS3Objects.ts
```diff
@@ -0,0 +1,26 @@
...
+  const command = new ListObjectsV2Command(input)
+  const response = await client.send(command)
+
+  if (response.Contents) {
```
SamBerry-GDS, line 15: We could reduce nested if statements here if we return the object when there's no content.

src/services/locateS3BucketData.ts
```diff
@@ -1,15 +1,108 @@
...
     glacierTierLocations: ['myGlacierLocation2'],
     dataAvailable: true
   })
 }
+
+export const getObjectPrefixes = (
+  dateFrom: string,
+  dateTo: string
+): string[] => {
+  const dateFromParts = dateFrom.split('/')
```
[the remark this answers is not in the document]
SamBerry-GDS: or rather a date to  UNIX epoch function

src/services/locateS3BucketData.ts
```diff
@@ -1,15 +1,108 @@
...
+
+  while (currentDate <= new Date(end)) {
+    dates.push(new Date(currentDate))
+    currentDate.setUTCHours(currentDate.getUTCHours() + 1)
+  }
+
+  // Include the first hour of the next day
+  dates.push(new Date(currentDate))
+
+  const objects = dates.map((date) => {
```
SamBerry-GDS, line 67: would a better name for this var be filePrefixes?

src/services/locateS3BucketData.ts
```diff
@@ -1,15 +1,108 @@
...
+    const year = date.getUTCFullYear()
+    const month = (date.getUTCMonth() + 1).toString().padStart(2, '0')
+    const day = date.getUTCDate().toString().padStart(2, '0')
+    const hour = date.getUTCHours().toString().padStart(2, '0')
+
```
SamBerry-GDS, line 72: This might be a bit simpler to read if these bits of logic were grouped in functions e.g. getEpochDates, validateEpochDates, generateDateRange and dateRangeToFilePrefix.

src/utils/helpers.ts
```diff
@@ -0,0 +1,8 @@
+export const getEnv = (name: string) => {
```
SamBerry-GDS, line 1: rather than the constants file above could we make this function take a string literal type with the expected env vars?  Something like:

`envVars: "ANALYSIS_BUCKET_NAME" | "AUDIT_BUCKET_NAME"
`

src/services/listS3Objects.ts
gdsandrewmccafferty: Yes, and actually you can just use `AWS_REGION` because I believe the Lambda runtime just provides it anyway. I was trying to find an official source for this, but check out e.g. this SO answer: https://stackoverflow.com/a/43095295

gdsandrewmccafferty (commented):
re: the change from `testUtils` to `utils`. Could we somehow put code that's just for tests somewhere else, so it's clearer? I guess you've now got common "production" code in there now, so you needed a place to put it, but I'd still like to find some way to make it clear that some code is just for tests, if we can.

src/services/listS3Objects.ts
andylangridge: Fixed.

src/services/listS3Objects.spec.ts
andylangridge: Have done so, can you let me know if it's descriptive enough?

`response has continuation token - return results for all pages`

src/services/locateS3BucketData.ts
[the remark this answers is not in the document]
andylangridge: Have added a helper function for this `getEpochDate()`, that naming ok?

src/services/locateS3BucketData.ts
andylangridge: As it now lives in a function called `timeRangeToObjectPrefixes()`, this is now called `timeRange` if that's ok?

src/services/locateS3BucketData.ts
andylangridge: Done, please check out the new function names!

src/services/locateS3BucketData.ts
```diff
@@ -1,15 +1,108 @@
...
+import { listS3Objects } from './listS3Objects'
+import { ANALYSIS_BUCKET_NAME, AUDIT_BUCKET_NAME } from '../utils/constants'
 
 export const locateS3BucketData = async (
   dataRequestParams: DataRequestParams
 ): Promise<S3BucketDataLocationResult> => {
   console.log('Looking for S3 data using params', dataRequestParams)
-  // TODO: fill in logic to search S3 bucket for data.
+
+  const objectPrefixes = getObjectPrefixes(
```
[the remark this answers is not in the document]
andylangridge: Fixed.

gdsandrewmccafferty:
> re: the change from `testUtils` to `utils`. Could we somehow put code that's just for tests somewhere else, so it's clearer? I guess you've now got common "production" code in there now, so you needed a place to put it, but I'd still like to find some way to make it clear that some code is just for tests, if we can.

Having looked at the code properly again, I can see that what you've done is fine!

src/utils/helpers.ts
andylangridge: Done, please review!

src/services/checkS3BucketData.ts
```diff
@@ -0,0 +1,56 @@
...
+  dataRequestParams: DataRequestParams
+): Promise<S3BucketDataLocationResult> => {
+  console.log('Looking for S3 data using params', dataRequestParams)
+
+  const prefixes = generateS3ObjectPrefixes(
+    dataRequestParams.dateFrom,
+    dataRequestParams.dateTo
+  )
+
+  //TODO: add handling for when there is no data available for requested dates
```
gdsandrewmccafferty, line 17: Can't we do this now? Just by looking to see if there's anything there at all? If we're not doing this in this story, not sure where we are doing it?

mabon-gds (approved):

PR #31, opened 2022-08-31
Feature/update zendesk ticket

mabon-gds:


src/services/updateZendeskTicket.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+const generateTestValidAuth = () => {
+  mockBase64Encode.mockReturnValue('VALIDEncodedAuthValue')
+}
+const generateTestInvalidAuth = () => {
+  mockBase64Encode.mockReturnValue('INVALIDEncodedAuthValue')
+}
+const mockMakeHttpsRequest = makeHttpsRequest as jest.Mock<void>
+const givenSuccessfulApiCall = () => {
+  mockMakeHttpsRequest.mockImplementation(
+    (options: Record<string, any>, postData: Record<string, any>, cb) => {
```
gdsandrewmccafferty, line 43: I'm not a fan of abbreviations when we have space for longer names, so could we just use `callback` instead of `cb`?

src/services/updateZendeskTicket.ts
```diff
@@ -1,8 +1,46 @@
...
+) {
+  if (!eventBody) {
+    console.log('No Zendesk info available. Cannot update ticket.')
+    return
+  }
+  const zendeskTicketInfo = JSON.parse(eventBody)
+  const secrets = await retrieveZendeskApiSecrets()
+  const options: https.RequestOptions = {
+    method: 'PUT',
+    hostname: 'govuk1620731396.zendesk.com',
```
gdsandrewmccafferty, line 17: I guess we'll get to it, but make sure we put the hostname into an environment variable!

src/services/httpsRequestUtils.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { makeHttpsRequest, base64Encode } from './httpsRequestUtils'
+import { exampleEventBody } from '../testUtils/events/exampleEventBody'
+
+function getSecrets() {
+  return {
+    ZENDESK_USER_ID: process.env.ZENDESK_USER_ID ?? '', // TODO grab from AWS Secrets Manager
```
gdsandrewmccafferty, line 6: This stuff will need to come from the secrets retrieval code I've now merged into the main branch, should hopefully work without any messing about!

src/services/httpsRequestUtils.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { base64Encode } from './httpsRequestUtils'
+
+describe('#base64Encode', () => {
+  it('returns the correctly encoded value along with the "Basic " prefix', () => {
+    const value = 'helloWorld'
```
gdsandrewmccafferty, line 5: Might be better if you used something closer to what you'll actually be using in practice, which I think is a token and username or something? Obviously not a real token, but just something that has the same shape if you can?

src/services/updateZendeskTicket.ts
```diff
@@ -1,8 +1,41 @@
...
+  if (!eventBody) {
+    console.log('No Zendesk info available. Cannot update ticket.')
+    return
+  }
+  const zendeskTicketInfo = JSON.parse(eventBody)
+  const secrets = await retrieveZendeskApiSecrets()
+  const options: https.RequestOptions = {
+    method: 'PUT',
+    hostname: process.env.ZENDESK_HOSTNAME,
+    path: `/api/v2/tickets/${zendeskTicketInfo.zendeskId}.json`,
```
gdsandrewmccafferty, line 18: I think we should check for the absence of `zendeskId`, and if it's not there log that fact, and not make the API call. Remember we're dealing a potentially malformed request, so it could actually have this field missing.

src/services/httpsRequestUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function makeHttpsRequest(
+  options: https.RequestOptions,
+  postData: Record<string, unknown> | undefined = undefined
+) {
+  return new Promise((resolve, reject) => {
+    const req = https.request(options, (response) => {
+      console.log('STATUS: ' + response.statusCode)
+
+      if (
+        (!response && response['statusCode'] < 200) ||
```
SamBerry-GDS, line 12: Can this ever return true?  We're checking the response object is falsey and then trying to get the value of statusCode from it.

src/services/updateZendeskTicket.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    givenAllSecretsAvailable()
+  })
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('a single api call was made with valid authentication', async () => {
+    generateTestValidAuth()
+    givenSuccessfulApiCall()
+    jest.spyOn(global.console, 'log')
```
SamBerry-GDS, line 63: Would it make sense to move this into your beforeEach hook?

src/services/httpsRequestUtils.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { base64Encode } from './httpsRequestUtils'
+
+describe('#base64Encode', () => {
```
rikumolugds, line 3: I'm not sure what the lambda expects, but I think making the request with at least one valid and one invalid request would make for interesting test cases. Suggestions:
> Missing headers, empty body?

src/services/updateZendeskTicket.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    return { theReturnData: '123' }
+  })
+}
+const givenUnsuccessfulApiCall = () => {
+  mockMakeHttpsRequest.mockImplementation(() => {
+    throw new Error('There was an error.')
+  })
+}
+
+describe('updating a zendesk ticket', () => {
```
rikumolugds, line 50: Suggested test cases:
> Missing ticket id
> Ticket already 'closed'
> Zendesk endpoint down

src/services/updateZendeskTicket.ts
```diff
@@ -1,8 +1,45 @@
...
+    headers: {
+      Authorization: base64Encode(
+        `${secrets.zendeskApiUserEmail}/token:${secrets.zendeskApiKey}`
+      ),
+      'Content-Type': 'application/json'
+    }
+  }
+  const postData = {
+    ticket: {
+      status: 'open',
```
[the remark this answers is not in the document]
mabon-gds: Ah yes! Thanks - added a configurable parameter so we can re-use the function to update tickets in different contexts

src/services/updateZendeskTicket.spec.ts
mabon-gds: done!

src/services/httpsRequestUtils.spec.ts
mabon-gds: now properly accessing secrets from AWS following your merge

src/services/httpsRequestUtils.spec.ts
mabon-gds: done!

src/services/updateZendeskTicket.ts
mabon-gds: added a check for ZendeskId along with a test

src/services/updateZendeskTicket.spec.ts
```diff
@@ -0,0 +1,134 @@
+// Under test
+import { updateZendeskTicket } from './updateZendeskTicket'
+// Dependencies
+import { makeHttpsRequest, base64Encode } from './httpsRequestUtils'
+import { retrieveZendeskApiSecrets } from './retrieveZendeskApiSecrets'
+import { ZendeskApiSecrets } from '../types/zendeskApiSecrets'
+import { exampleEventBody } from '../testUtils/events/exampleEventBody'
+
+// Mocking retrieveZendeskApiSecrets
```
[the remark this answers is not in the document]
mabon-gds: done!

src/services/updateZendeskTicket.spec.ts
mabon-gds: done!

src/services/updateZendeskTicket.spec.ts
mabon-gds: addressed

SamBerry-GDS (approved):

PR #38, opened 2022-09-02
Feature/tt2 11  params validation

gdsandrewmccafferty:


src/services/generateS3ObjectPrefixes.spec.ts
```diff
@@ -123,19 +123,19 @@ describe('object prefixes', () => {
...
-    const result = generateS3ObjectPrefixes('2022/11/21', '2022/11/21')
+    const result = generateS3ObjectPrefixes('2022-11-21', '2022-11-21')
     expect(result).toEqual(expectedResult)
   })
 
   test('invalid date string', () => {
     expect(() => {
       generateS3ObjectPrefixes('invalid', 'invalid')
-    }).toThrow('String not valid date')
+    }).toThrow("String 'invalid' is not a valid date")
```
andylangridge, line 133: This will fail the pipeline won't it? I think Linting will want single quotes.

gdsandrewmccafferty:
I'm replying this on my phone, let's see if it ends up on the PR! Isn't
there a special case in the linting for when there are single quotes inside
a double-quoted string (which there is here)

src/services/validateZendeskRequest.ts
```diff
@@ -1,17 +1,158 @@
...
+    },
+    {
+      message: 'At least one journey id should be provided',
+      isValid:
+        data.identifierType != 'journey_id' || data.journeyIds?.length > 0
+    },
+    {
+      message: 'At least one event id should be provided',
+      isValid: data.identifierType != 'event_id' || data.eventIds?.length > 0
+    },
```
rikumolugds, line 58: There is the  `User ID` identifier also. Should we be checking for that as well?

src/services/validateZendeskRequest.spec.ts
```diff
@@ -0,0 +1,412 @@
...
+      JSON.stringify(buildValidRequestBodyWithIds('event_id', 'id1 id2 id3'))
+    )
+
+    expect(validationResult.dataRequestParams?.eventIds).toEqual([
+      'id1',
+      'id2',
+      'id3'
+    ])
+  })
+
```
rikumolugds, line 242: There is the `User ID` identifier also on the Zendesk form. Should we be checking for that as well?

src/services/validateZendeskRequest.spec.ts
```diff
@@ -0,0 +1,412 @@
+import { validateZendeskRequest } from './validateZendeskRequest'
+import { IdentifierTypes } from '../types/dataRequestParams'
+describe('validateZendeskRequest', () => {
```
rikumolugds, line 3: A very corner case, but could be expected since we can't remove the attachment field. If the form has an attachment, would that be submitted along, and if so, do we just ignore it or send Zendesk some message? Not sure if the Zendesk macros would have already discarded that.

src/services/generateS3ObjectPrefixes.spec.ts
andylangridge: Having checked, looks like Prettier is ok with this due to the quotes within quotes!

src/services/validateZendeskRequest.spec.ts
andylangridge: Don't know much about the implementation, but I assume what you see in Zendesk is not what is sent via the webhook. We are in control of what we send to the API, so we don't have to send every field from Zendesk.

Maybe @mabon-gds can confirm in Andrew's absence?

src/services/validateZendeskRequest.spec.ts
```diff
@@ -0,0 +1,412 @@
...
+    expect(validationResult.dataRequestParams?.zendeskId).toEqual(testZendeskId)
+    expect(validationResult.dataRequestParams?.resultsEmail).toEqual(
+      testValidResultsEmail
+    )
+    expect(validationResult.dataRequestParams?.resultsName).toEqual(
+      testResultsName
+    )
+    expect(validationResult.dataRequestParams?.identifierType).toEqual(
+      'session_id'
+    )
```
rikumolugds, line 151: I see the next TC checks for specific values, but I'd suggest we at least check the list is not empty at this point.

src/services/validateZendeskRequest.spec.ts
kasgds: I think we should be able to remove the attachment field as I remember seeing that a template. I'll investigate this and action it separately.

src/services/validateZendeskRequest.spec.ts
```diff
@@ -0,0 +1,412 @@
+import { validateZendeskRequest } from './validateZendeskRequest'
```
rikumolugds, line 1: Final suggestions:
> `validateZendeskRequest.ts` does not seem to validate rule 1 (check all fields are present) as per the ticket. Can we add that? 
> Check date values are not in the future.

src/services/validateZendeskRequest.spec.ts
kasgds: I can't see an easy way to remove the field as I previously commented but back to the original concern. Andy is right here, we only send the fields we're concerned with so even if someone attaches something to the form, by design, we're not including it in the webhook. The same goes for Subject and Description fields both of which are system fields. You can see the details [here](https://govuk1620731396.zendesk.com/admin/objects-rules/rules/triggers/5711827450012) of what's sent here.

src/services/validateZendeskRequest.spec.ts
kasgds: This has been added and ready for review.

kasgds: > There is the `User ID` identifier also on the Zendesk form. Should we be checking for that as well?

This has been added

src/services/validateZendeskRequest.spec.ts
rikumolugds: Thanks for clarifying.

src/services/validateZendeskRequest.spec.ts
kasgds: I've added a test to check the existence of all fields and we already have a tests in place for future dates.

rikumolugds (dismissed):

rikumolugds (approved):

rikumolugds (dismissed):

rikumolugds (approved):

PR #41, opened 2022-09-02
replaced ZendeskHostName with a secret

mabon-gds:
ZendeskHostName is now accessed from secretsManager


andylangridge (changes requested):
We'll also need to remove the parameter from the SAM template!

andylangridge (approved):

PR #43, opened 2022-09-05
Add script to sync buckets

andylangridge:
Script will sync the staging audit bucket to a target bucket name. Allows for user input for the target, but a default value of the sandbox bucket is provided. Script assumes required admin roles for each account are configured in your AWS CLI config.


Aidan23Smith (commented):
Just a quick comment, I don't think we should have our AWS account numbers in the file (but that depends if it will be public or not)

gds-topickering (approved):

Aidan23Smith (approved):
Looks good to me

PR #48, opened 2022-09-06
Feature/tt2 26  athena data source

gds-topickering:
Create Athena resources via SAM template


tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "noImplicitAny": true,
-    "lib": ["es2022"],
+    "lib": ["es2022", "dom"],
```
SamBerry-GDS, line 8: I hadn't thought we'd need dom as we're not running our code in a browser.  Is this an Athena requirement?

gds-topickering: Good spot, this was a change I had to put in to stop VSCode constantly flagging errors, I will take it out and check if the errors persist

gds-topickering: Addressed by removing dom

SamBerry-GDS (dismissed):

template.yaml
```diff
@@ -274,3 +274,71 @@ Resources:
...
+            Parameters: {"ignore.malformed.json": true,"serialiazation.format": 1,"field.delim":""}
+            SerializationLibrary: "org.openx.data.jsonserde.JsonSerDe"
+          StoredAsSubDirectories: false
+        TableType: EXTERNAL_TABLE
+
+  AthenaAnalysisWorkgroup:
+    Type: AWS::Athena::WorkGroup
+    Properties:
+      Description: The workgroup for TxMA queries
+      Name: TxMAAnalysisQueries
```
andylangridge, line 332: Same question here re conflicts!

template.yaml
```diff
@@ -274,3 +274,71 @@ Resources:
...
+      WorkGroupConfiguration:
+        EnforceWorkGroupConfiguration: true
+        PublishCloudWatchMetricsEnabled: false
+        ResultConfiguration:
+          EncryptionConfiguration:
+            EncryptionOption: "SSE_S3"
+          OutputLocation:
+            !Join
```
andylangridge, line 340: Else where we've used the shorthand way of writing these expressions, which personally I find a bit easier to read!

This could be `!Join [ '' [' s3://', !Ref QueryResultsBucket, '// ] ]`

template.yaml
```diff
@@ -274,3 +274,71 @@ Resources:
...
         Status: Enabled
+
+  # Athena Data Source
+  AuditAnalysisDatabase:
+    Type: AWS::Glue::Database
+    Properties:
+      CatalogId: !Ref AWS::AccountId
+      DatabaseInput:
+        Description: "TxMA - Event Message Analysis Database"
+        Name: "txma-analysis"
```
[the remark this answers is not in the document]
gds-topickering: Potentially - to avoid the risk I will amend the name to follow the naming conventions elsewhere in the template

template.yaml
gds-topickering: See above - amending name to follow the convention

template.yaml
gds-topickering: Changed as per suggestion

andylangridge (dismissed):

andylangridge (approved):

PR #49, opened 2022-09-07
Feature/tt2 16 ensure only zendesk can start process

mabon-gds:
this PR comprises of work done to meet the ACs of TT2-16. 
Additionally, commit number `30c7c36b9c23517bdb734703e897defd75d08c63` includes a refactor of  2 test suites where mocking code was duplicated. If anyone knows of a cleaner way to deal with using similar mocks across test suites, please let me know and I'll be happy to change it up!


src/handlers/initiateDataRequest.ts
```diff
@@ -1,18 +1,28 @@
...
 
 export const handler = async (
   event: APIGatewayProxyEvent
 ): Promise<APIGatewayProxyResult> => {
+  const headerSignature = event.headers['X-Zendesk-Webhook-Signature'] as string
+  const timestamp = event.headers[
+    'X-Zendesk-Webhook-Signature-Timestamp'
+  ] as string
+  if (
+    !(await isValidSignature(headerSignature, event.body as string, timestamp))
```
SamBerry-GDS, line 17: If this is the only place we use the function it might be better to call it `isSignatureInvalid` and return true if it's not valid.  That way we don't need the not operator.

SamBerry-GDS: I'm also not sure about casting the body and headers to strings - could they be undefined, is that something we should handle in the function where we validate the signature?

src/handlers/initiateDataRequest.spec.ts
```diff
@@ -80,10 +100,25 @@ describe('initate data request handler', () => {
     expect(initiateDataTransfer).toHaveBeenCalledWith(testDataRequest)
   })
 
-  it('returns 400 response when request is invalid', async () => {
+  it('returns 400 response when request signature is invalid', async () => {
+    jest.spyOn(global.console, 'warn')
+    givenSignatureIsInvalid()
+    expect(await callHandlerWithBody()).toEqual({
+      statusCode: 400,
```
SamBerry-GDS, line 107: It might be nice to break the execution out of the assertion (we do it in a few places in this file) and separate the setup, execution and assertion with an empty line.

src/handlers/initiateDataRequest.ts
```diff
@@ -1,18 +1,28 @@
...
+  ) {
+    return await handleInvalidRequest()
```
gdsandrewmccafferty, line 19: Since you're doing something very different in the two cases, I would just add another method to handle the case of the signature being invalid. `handleInvalidRequest` has become a bit confusing now, so that would help with that too.

gdsandrewmccafferty: I also wonder if to avoid clutter here, it would be easier if `isValidSignature` just accepted an `APIGatewayProxyEvent` and you can then handle all the stuff around pulling the headers out and checking they're set, and whether the body is missing or not in that function, rather than inflating the code in this handler.

src/handlers/initiateDataRequest.ts
mabon-gds: addressed in latest commit. Please review

src/handlers/initiateDataRequest.spec.ts
mabon-gds: changed the code to adhere to this format in latest commit

src/handlers/initiateDataRequest.ts
mabon-gds: have addressed these in latest commit. Please review

src/handlers/initiateDataRequest.ts
```diff
@@ -44,3 +48,14 @@ const handleInvalidRequest = async (
     })
   }
 }
+
+const handleInvalidSignature = async () => {
+  console.warn('Request received with invalid webhook signature')
+  const validationMessage = 'Invalid request source'
```
gdsandrewmccafferty, line 54: I would just inline this variable, it's not adding much.

gdsandrewmccafferty (approved):

PR #50, opened 2022-09-07
Custom resource to clean up S3 Buckets

andylangridge:


src/services/listS3ObjectVersions.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    })
+
+    const result = await listS3ObjectVersions(input)
+    expect(result).toEqual({
+      deleteMarkers: ['version-1'],
+      versions: ['version-1']
+    })
+  })
+
+  test('respnse has NextKeyMarker', async () => {
```
gdsandrewmccafferty, line 58: Just for neatness, there's a typo a few times here that says `respnse`

src/services/emptyS3Bucket.ts
```diff
@@ -0,0 +1,59 @@
...
+  GetBucketVersioningCommand,
+  PutBucketVersioningCommand,
+  S3Client
+} from '@aws-sdk/client-s3'
+import { getEnv } from '../utils/helpers'
+import { listS3Objects } from './listS3Objects'
+import { listS3ObjectVersions } from './listS3ObjectVersions'
+
+export const emptyS3Bucket = async (bucketName: string): Promise<void> => {
+  const s3Client = new S3Client({ region: getEnv('AWS_REGION') })
```
gdsandrewmccafferty, line 12: This function looks a bit long to me, could it be split up? Also, there's a very long block inside an `if`, wonder if that could be made clearer?

gdsandrewmccafferty (approved):

PR #56, opened 2022-09-15
Check for Athena Data Source

gds-topickering:


package.json
```diff
@@ -19,6 +19,7 @@
     "test:cov": "jest --coverage"
   },
   "dependencies": {
+    "@aws-sdk/client-glue": "^3.170.0",
```
SamBerry-GDS, line 22: We can't do variable version installs.  I'd also keep this at the same version as the other aws-sdk packages.

gds-topickering: Will amend to 3.154.00

template.yaml
```diff
@@ -354,7 +355,69 @@ Resources:
       VersioningConfiguration:
         Status: Enabled
 
-  # Query Bucket
+  # Data Analysis
+
+  InitiateAthenaQueryFunction:
+    #checkov:skip=CKV_AWS_116:Lambda is behind an API gateway, so repeated processing of the same message isn't a problem
```
gdsandrewmccafferty, line 361: In this case, I think you do want a dead-letter queue, because you're processing stuff off a queue. Have a look at what's now in the main branch, for ProcessDataRequestFunction

gds-topickering: Thanks, have removed the skip and have added a dead letter queue

gdsandrewmccafferty (approved):

PR #57, opened 2022-09-15
Feature/tt61 sqs between validation and data retrieval

gdsandrewmccafferty:


src/handlers/initiateDataRequest.spec.ts
```diff
@@ -34,7 +29,15 @@ jest.mock('../services/validateRequestSource', () => ({
   isSignatureInvalid: jest.fn()
 }))
 
+jest.mock('../services/queue/sendInitiateDataTransferMessage', () => ({
+  sendInitiateDataTransferMessage: jest.fn()
+}))
+
 describe('initate data request handler', () => {
+  beforeEach(() => {
```
SamBerry-GDS, line 37: Purely a preference thing but I prefer the hooks just before the tests and after the vars.

src/handlers/processDataRequest.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  it('should handle a valid data request event', async () => {
+    const initiateDataRequestEvent = constructInitiateDataRequestEvent(
+      JSON.stringify(testDataRequest)
+    )
+    await handler(initiateDataRequestEvent)
+    expect(initiateDataTransferMock).toHaveBeenCalledWith(testDataRequest)
+  })
+
+  it('should throw an appropriate error if there is no data in the event', () => {
+    expect(handler({ Records: [] })).rejects.toThrow('No data in event')
```
SamBerry-GDS, line 28: This needs to be async - the next line is probably passing before this runs.

src/handlers/processDataRequest.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    expect(initiateDataTransferMock).not.toHaveBeenCalled()
+  })
+
+  it('should throw an appropriate error if the request includes data of the wrong shape', async () => {
+    const initiateDataRequestEvent = constructInitiateDataRequestEvent(
+      JSON.stringify({ someProperty: 'someValue' })
+    )
+    expect(handler(initiateDataRequestEvent)).rejects.toThrow(
```
SamBerry-GDS, line 36: I think we need an await before the handler on this line for the next assertion to do anything

src/services/queue/sendInitiateDataTransferMessage.ts
```diff
@@ -0,0 +1,11 @@
+import { DataRequestParams } from '../../types/dataRequestParams'
+import { getEnv } from '../../utils/helpers'
+import { sendSqsMessage } from './sendSqsMessage'
+export const sendInitiateDataTransferMessage = async (
+  dataRequestParams: DataRequestParams
+) => {
+  await sendSqsMessage(
```
SamBerry-GDS, line 7: Do we need to call await here or can we fire and forget?

Would it make more sense if this function returned the promise so other functions could wait on it's completion?

src/utils/tests/events/initiateDataRequestEvent.ts
```diff
@@ -0,0 +1,25 @@
...
+          'AROAWHRG3MZTWIX3SW3BQ:tt61-sqs-between-validati-InitiateDataRequestFunct-jYG1YRNmEet2',
+        ApproximateFirstReceiveTimestamp: '1663168135670'
+      },
+      messageAttributes: {},
+      md5OfBody: '80f5f39c0a260622d3313a3040420a58',
+      eventSource: 'aws:sqs',
+      eventSourceARN:
+        'arn:aws:sqs:eu-west-2:428504475239:tt61-sqs-between-validation-and--InitiateDataRequestQueue-C7dEDDYuULrB',
+      awsRegion: 'eu-west-2'
+    }
```
gds-topickering, line 23: Are there any concerns with these being genuine values?

src/services/queue/sendSqsMessage.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  SQSClient,
+  SendMessageRequest,
+  SendMessageCommand
+} from '@aws-sdk/client-sqs'
+import { getEnv } from '../../utils/helpers'
+
+export const sendSqsMessage = async (messageBody: object, queueUrl: string) => {
```
gds-topickering, line 8: Is it possible to be more specific about the messageBody type or is this something that might be expanded later?

src/handlers/processDataRequest.spec.ts
gdsandrewmccafferty: Yep, good point. Originally I had in inside the expect, but it turns out the way to do is to put it outside, like this https://stackoverflow.com/a/47887098

src/services/queue/sendSqsMessage.ts
gdsandrewmccafferty: Yep, the idea behind this code is that it's completely generic, we could put anything in here (and almost certainly will). We could have written a really specific bit of code to send the message type we wanted to send for our case, but I knew there were more coming, so to avoid future people some boilerplate, this is what we came up with.

src/services/queue/sendInitiateDataTransferMessage.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+import { sendSqsMessage } from './sendSqsMessage'
+jest.mock('../../utils/helpers', () => ({
+  getEnv: jest.fn()
+}))
+jest.mock('./sendSqsMessage', () => ({
+  sendSqsMessage: jest.fn()
+}))
+
+const mockGetEnv = getEnv as jest.Mock<string>
+const mockSendSqsMessage = sendSqsMessage as jest.Mock
```
gds-topickering: You can use the utils/tests/setup/testEnvVars for the URL, rather than mocking it

src/utils/tests/events/initiateDataRequestEvent.ts
gdsandrewmccafferty: Good point. I don't think these are very sensitive values, but I've made them look more fake now. If these had been tokens or something we'd have had to regenerate them, but as it is, I think this is all temporary stuff that was generated one-off in this stack, and isn't something anybody can do anything with.

src/services/queue/sendInitiateDataTransferMessage.spec.ts
gdsandrewmccafferty: Ah yep, hadn't noticed that! I'm sort of in two minds about centralising that kind of thing, but since it's there I may as well use it like that!

SamBerry-GDS (approved):

src/handlers/processDataRequest.spec.ts
gdsandrewmccafferty: Done!

gds-topickering (approved):

PR #58, opened 2022-09-15
Test error response when zendesk signature invalid

SamBerry-GDS:
Adds test an error is returned for invalid signature calls for TT2-16


integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -0,0 +1,50 @@
+import axios from 'axios'
+
+describe('Zendesk request API', () => {
+  it('returns a Invalid request source on invalid Zendesk Webhook Signature', async () => {
```
rikumolugds, line 4: `API Gateway returns an Invalid request source when request has invalid Zendesk Webhook Signature`?

integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -0,0 +1,50 @@
+import axios from 'axios'
+
+describe('Zendesk request API', () => {
```
rikumolugds, line 3: `Zendesk request integrity` perhaps?

integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const invalidRequestError = async () => {
+      return axios({
+        url: `https://${awsBaseUrl}/Stage/zendesk-webhook`,
+        method: 'POST',
+        headers,
+        data
+      })
+        .then(() => {
+          throw 'Axios did not recieve an error'
```
rikumolugds, line 36: `Gateway did not return an error` just to make it client agnostic?

integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+        })
+        .catch((error) => {
+          if (error.response) {
+            return error.response
+          }
+          throw 'Axios did not recieve a response error'
```
rikumolugds, line 42: Same as above.

integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+        })
+    }
+
+    const errorResponse = await invalidRequestError()
+    expect(errorResponse.status).toBe(400)
```
rikumolugds, line 47: The acceptance criteria states unauthorised, suggesting a 401. Do we need to update the a/c? Also, there's a final a/c about the attempt being logged, not sure if that should be in Cloudwatch?

SamBerry-GDS: We could - a 401 suggests you should re-authenticate which isn't how this works though.  I think a 400 is probably right for whats actually happening in this case.

rikumolugds (approved):

PR #61, opened 2022-09-20
Feature/tt2 13 check data storage tier

gdsandrewmccafferty:
This PR adds a step to close any Zendesk tickets where there's no available data for that time period, and also figures out which files that need to be copied are in Glacier storage


src/services/checkS3BucketData.ts
```diff
@@ -14,43 +15,58 @@ export const checkS3BucketData = async (
...
   const objectsToCopy = requestedAuditBucketObjects.filter(
-    (object) => !existingAnalysisBucketObjects.includes(object)
+    (object) =>
+      !existingAnalysisBucketObjects.map((o) => o.Key).includes(object.Key)
   )
 
-  console.log('Objects to copy:', objectsToCopy)
+  const standardTierLocationsToCopy = objectsToCopy
+    .filter((o) => o.StorageClass === 'STANDARD')
+    .map((o) => o.Key as string)
```
SamBerry-GDS, line 41: Looks like Key and StorageClass could be undefined: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/_object.html

The filter should be fine but is it worth checking we aren't adding undefined values into the final lists

gdsandrewmccafferty: Yeah, I was confused as to how you could have an item in a bucket without a key, but I guess for safety we could filter out anything that doesn't have either of those properties set.

gdsandrewmccafferty: Have sorted this now, added some filtering for any weird objects like this in our buckets

src/services/checkS3BucketData.spec.ts
```diff
@@ -120,20 +161,129 @@ describe('check objects in analysis bucket', () => {
...
-  test('partial data in analysis bucket', async () => {
-    givenPartialDataInAnalysisBucket(prefixes)
+  test('no data in analysis bucket, audit bucket data contains some data with missing keys', async () => {
+    givenDataInBucketForPrefixes(
+      [prefixes[0], prefixes[2]],
+      TEST_AUDIT_BUCKET,
+      'STANDARD'
+    )
+    givenDataInBucketForPrefix(prefixes[1], TEST_AUDIT_BUCKET, [
+      { StorageClass: 'STANDARD' }
```
SamBerry-GDS, line 174: Is it also worth adding an object with `{ StorageClass: 'STANDARD', Key: undefined }`

SamBerry-GDS (dismissed):

src/services/checkS3BucketData.spec.ts
gdsandrewmccafferty: Done

SamBerry-GDS (approved):

PR #63, opened 2022-09-22
Feature/add dynamo db

SamBerry-GDS:
Adds infrastructure for DynamoDB


template.yaml
```diff
@@ -299,6 +330,12 @@ Resources:
                 - kms:Decrypt
               Resource:
                 - !GetAtt QueueKmsKey.Arn
+            - Sid: DynamoDBPut
+              Effect: Allow
+              Action:
+                - dynamodb:PutItem
```
gdsandrewmccafferty, line 336: Just wondering if we might need to query as well from the ProcessDataRequest function? Things like "is there already processing happening on this ticket"? We can add it later I guess though, so this is more of a point for discussion.

gdsandrewmccafferty (approved):

template.yaml
SamBerry-GDS: It's possible, I'd rather add it if we definitely need it though.

PR #64, opened 2022-09-23
Feature/tt2 21  notify integration

mabon-gds:
Should note that the event type is a placeholder at the moment as we’re currently unsure what’s going to sit in between this lambda and the previous. This also means that there’s no trigger for this lambda yet, so will need to test manually on AWS to see it working.


src/handlers/sendEmailRequestToNotify.ts
```diff
@@ -0,0 +1,41 @@
+import { APIGatewayProxyEvent } from 'aws-lambda'
+import { NotifyClient } from 'notifications-node-client'
+import { retrieveNotifySecrets } from '../secrets/retrieveNotifySecrets'
+import { CustomAxiosResponse } from '../types/notify/axiosResponse'
+import { tryParseJSON } from '../utils/helpers'
+
+// event type is a placeholder
+export const handler = async (event: APIGatewayProxyEvent) => {
+  if (!event.body) {
+    console.log('Could not find event body. An email has not been sent')
```
gdsandrewmccafferty, line 10: I think this should probably be `console.error` because if this happens it indicates something wrong in the caller, so we'll want to know about it. I'm assuming at some point we'll be able to filter errors in our logs, and do things like get notifications when we get a lot of them.

src/handlers/sendEmailRequestToNotify.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+        }
+      }
+    )
+    expect(console.log).toHaveBeenCalledWith({
+      status: 201,
+      emailSentTo: TEST_NOTIFY_EMAIL,
+      subjectLine: 'Your data query has completed'
+    })
+  })
+  xit('returns a 400 status code with a message', async () => {
```
gdsandrewmccafferty, line 81: You've skipped this one with `xit` for some reason?

src/handlers/sendEmailRequestToNotify.ts
```diff
@@ -0,0 +1,41 @@
...
+  try {
+    const response = (await Promise.resolve(
+      notifyClient.sendEmail(secrets.notifyTemplateId, requestDetails.email, {
+        personalisation: {
+          firstName: requestDetails.firstName,
+          zendeskId: requestDetails.zendeskId,
+          signedUrl: requestDetails.signedUrl
+        }
+      })
+    )) as unknown as CustomAxiosResponse
```
gdsandrewmccafferty, line 28: Not a huge fan of this double-cast, wonder if there's a way to do it without this?

src/handlers/sendEmailRequestToNotify.ts
```diff
@@ -0,0 +1,41 @@
...
+
+// event type is a placeholder
+export const handler = async (event: APIGatewayProxyEvent) => {
+  if (!event.body) {
+    console.log('Could not find event body. An email has not been sent')
+    return
+  }
+
+  const requestDetails = tryParseJSON(event.body)
+  const secrets = await retrieveNotifySecrets()
```
gdsandrewmccafferty, line 15: I'd like to see something here to check that the required fields are set in the SQS event body (and throw an error if they're not). That way, we don't get weird unexpected errors further down the line from things being missing.

src/handlers/sendEmailRequestToNotify.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    await callHandlerWithBody(invalidEventBody)
+
+    expect(console.error).toHaveBeenCalledWith(
+      'Could not find event body. An email has not been sent'
+    )
+    expect(NotifyClient).not.toHaveBeenCalled()
+  })
+  it('returns from the function and logs an error when a value is missing from the event body', async () => {
+    const invalidEventBody = `{
```
gdsandrewmccafferty, line 120: I know it's a pain, but it would be good to test the case of each property missing, rather than just one.
You can make this very slightly neater by using an array of property names, and using `map` to run the test for each.

gdsandrewmccafferty (dismissed):

src/handlers/sendEmailRequestToNotify.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    const zendeskTicketUpdateComment =
+      'A link to your results has been sent to you.'
+    await closeZendeskTicket(
+      requestDetails.zendeskId,
+      zendeskTicketUpdateComment
+    )
+  } catch (error) {
+    console.error('There was an error sending a request to Notify: ', error)
+    if (requestDetails.zendeskId) {
```
rikumolugds, line 48: Wondering if we need this check here since line 16 has already checked this.

gdsandrewmccafferty (approved):

gdsandrewmccafferty (approved):

PR #68, opened 2022-09-27
Add initial logic to compare zendesk ticket with request

andylangridge:


src/services/matchZendeskTicket.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+  test('ticket and request match', async () => {
+    givenZendeskTicketMatches()
+
+    const matchResult = await matchZendeskTicket(testDataRequest)
+
+    expect(matchResult).toEqual(true)
+  })
+
+  test('ticket and request do not match', async () => {
+    givenZendeskTicketDoesNotMatch()
```
gdsandrewmccafferty, line 199: How have you selected which case to check for here? Given that we have code to check each field, ideally you'd want to test each one. You can probably find a parameterised way to do this to avoid repetition, but for full test coverage I think you'll need to do that.

src/services/matchZendeskTicket.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    return field
+  }
+
+  return assert(ticketDetails.custom_fields.find((field) => field.id === id))
+    .value as string
+}
+
+const compareTicketAndRequestParam = (
+  ticketParam: string | string[] | null,
```
gdsandrewmccafferty, line 41: I think this function is doing too much, and makes working out what's going on quite confusing. I think you'd be better off making methods for different types, and then you can just call the appropriate one for the field you're looking at.

src/services/getZendeskTicket.ts
```diff
@@ -0,0 +1,32 @@
...
+    headers: {
+      Authorization: base64Encode(
+        `${secrets.zendeskApiUserEmail}/token:${secrets.zendeskApiKey}`
+      )
+    }
+  }
+
+  const data = await makeHttpsRequest(options)
+
+  if (isZendesktTicketResult(data)) {
```
gdsandrewmccafferty, line 26: What happens if there isn't a ticket at all for the given id? I think the Zendesk api returns a 404 in that case, which I believe at the moment will cause our makeHttpsRequest function throw an error. You might need to tweak that to make this work.

andylangridge: I'm currently handling this in the handler....

```
try {
    if (!(await matchZendeskTicket(requestParams))) {
      return await handleUnmatchedRequest(requestParams)
    }
  } catch (error) {
    console.error(error)
    return {
      statusCode: 404,
      body: JSON.stringify({
        message: 'Zendesk ticket not found'
      })
    }
  }

I'm basically assuming that ANY error we get back from the `matchZendeskTicket` function means we stop the process and say the ticket couldn't be found.

Possible errors we could get are errors from the Zendesk API calls (Show Ticket and Show User). One of which would be the legitimate 404 for the Show Ticket call. Do you think we should handle each of these cases individually? Trouble was I couldn't find anything to do with errors in their docs.... just the 200 response.

There is also the possibility that the custom field could not be found, so maybe we also handle that separately.

File: src/services/matchZendeskTicket.ts (hunk 2 above)
Comment (andylangridge) [reply src/services/matchZendeskTicket.ts]: Done!

File: src/handlers/initiateDataRequest.ts [hunk 4]
```diff
@@ -15,13 +20,29 @@ export const handler = async (
...
     return await handleInvalidRequest(event.body, validatedZendeskRequest)
   }
 
-  const messageId = await sendInitiateDataTransferMessage(
+  const requestParams =
     validatedZendeskRequest.dataRequestParams as DataRequestParams
-  )
+
+  try {
+    if (!(await matchZendeskTicket(requestParams))) {
```
SamBerry-GDS, line 32: As we'll only use this function here I'd switch the boolean response around so you don't need to reverse it in the if block.  Renaming it something like zendeskTicketDiffersToRequest lets you make this a bit simpler.

src/handlers/initiateDataRequest.ts
```diff
@@ -63,3 +84,20 @@ const handleInvalidSignature = async () => {
     })
   }
 }
+
+const handleUnmatchedRequest = async (requestDetails: DataRequestParams) => {
```
SamBerry-GDS, line 88: Would it be tidier to just pass in the zendeskId rather than the whole requestParam object?

src/services/getZendeskTicket.ts
```diff
@@ -0,0 +1,32 @@
...
+  isZendesktTicketResult,
+  ZendeskTicket
+} from '../types/zendeskTicketResult'
+import { base64Encode, makeHttpsRequest } from './httpsRequestUtils'
+import { retrieveZendeskApiSecrets } from './retrieveZendeskApiSecrets'
+import https from 'node:https'
+
+export const getZendeskTicket = async (
+  id: string
+): Promise<ZendeskTicket | undefined> => {
```
SamBerry-GDS, line 11: When you use this you casting it to ZendeskTicket so it would be better if this function couldn't return undefined.  Would it make sense to throw an error if ticketInfo can't be caught?

src/utils/tests/testDataRequest.ts
```diff
@@ -15,3 +15,27 @@ export const testDataRequest = {
   eventIds: ['123', '456'],
   piiTypes: ['passport_number']
 } as DataRequestParams
+
+export const testDataRequestWithValues = {
```
SamBerry-GDS, line 19: would testDataRequestWithAllValues be more descriptive (to differentiate it from the var above)?

src/handlers/initiateDataRequest.ts
gdsandrewmccafferty: Done!

src/services/getZendeskTicket.ts
gdsandrewmccafferty: Done!

src/utils/tests/testDataRequest.ts
gdsandrewmccafferty: Yep, have done that now

SamBerry-GDS (dismissed):

integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -1,12 +1,85 @@
 import axios from 'axios'
-import { getEnvVariable } from './lib/zendeskParameters'
+import crypto from 'crypto'
+
+//TODO: test setup and teardown for creating and deleting a ticket to use
+
+const baseUrl = process.env.ZENDESK_WEBHOOK_API_BASE_URL as string
```
rikumolugds, line 6: If you merge in main, there's already an env variable 'AWS_WEBHOOK_URL' that holds this.

integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -21,30 +94,39 @@ describe('Zendesk request integrity', () => {
...
-        })
+    const errorResponse = await invalidRequestError(headers, webhookRequestData)
+    expect(errorResponse.status).toEqual(404)
+    expect(errorResponse.data.message).toEqual('Zendesk ticket not found')
+  })
+
+  test('API Gateway returns a 400 response if the request does not match info in corresponding Zendesk ticket', async () => {
+    const webhookRequestData = {
+      //ticket with ID 881 exists with different info
+      zendeskId: '881',
```
rikumolugds, line 108: Could we maybe call the function to raise a ticket and use the id, rather than hardcode. The ticket could be deleted by anybody even outside our team. I'll add something to delete tickets shortly, so it can be used for teardown.

integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -21,30 +94,39 @@ describe('Zendesk request integrity', () => {
...
+
+    const headers = {
+      ...generateSignatureHeaders(webhookRequestData)
     }
 
-    const errorResponse = await invalidRequestError()
-    expect(errorResponse.status).toBe(400)
-    expect(errorResponse.data.message).toBe('Invalid request source')
+    const errorResponse = await invalidRequestError(headers, webhookRequestData)
+    expect(errorResponse.status).toEqual(400)
```
rikumolugds, line 127: Can you add a check for the log error as well as per the a/c? There are now helper functions for that.

src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.ts
```diff
@@ -0,0 +1,147 @@
...
+import {
+  getEnvAsNumber,
+  mapSpaceSeparatedStringToList
+} from '../../utils/helpers'
+
+export const zendeskTicketDiffersFromRequest = async (
+  requestParams: DataRequestParams
+) => {
+  console.log('Matching received request with existing Zendesk Tickets')
+  const ticketDetails = await getZendeskTicket(requestParams.zendeskId)
```
rikumolugds, line 15: Should we be anticipating a potential error here if the ticket does not exist?

src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+jest.mock('../../sharedServices/zendesk/getZendeskUser', () => ({
+  getZendeskUser: jest.fn()
+}))
+
+const mockGetZendeskTicket = getZendeskTicket as jest.Mock<
+  Promise<ZendeskTicket>
+>
+const mockGetZendeskUser = getZendeskUser as jest.Mock<Promise<ZendeskUser>>
+
+describe('match zendesk ticket details', () => {
```
rikumolugds, line 40: We are missing the scenario where the ticket does not exist. Could you please add that?

integration-tests/zendeskRequestAPI.spec.ts
andylangridge: I changed it on purpose because we didn't understand what it was at first. I think this makes it a bit clearer?

src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.ts
```diff
@@ -0,0 +1,147 @@
...
+    unmatchedParameters.push('userIds')
+
+  if (unmatchedParameters.length > 0) {
+    console.warn(
+      'Request does not match values on Ticket, the following parameters do not match:',
+      unmatchedParameters
+    )
+    return true
+  } else {
+    return false
```
rikumolugds, line 145: Should we be logging that the ticket info matches here to complete scenario 1?

integration-tests/zendeskRequestAPI.spec.ts
andylangridge: I'd rather not extend this PR any further (it's been open a long time now), so would look to pick that up separately.

integration-tests/zendeskRequestAPI.spec.ts
andylangridge: > This does the same thing though? Essentially we'd be calling AWS to look in the logs for this very message!

Same here!

rikumolugds (commented):
We might be missing the outstanding acceptance criteria for scenario 1, which would prevent us from adding tests for that.

integration-tests/zendeskRequestAPI.spec.ts
andylangridge: Also we will be adding more API gateways so we need to make a distinction between them

integration-tests/zendeskRequestAPI.spec.ts
```diff
@@ -1,12 +1,85 @@
...
+      .createHmac('sha256', process.env.ZENDESK_WEBHOOK_SECRET_KEY as string)
+      .update(timestamp + JSON.stringify(requestData))
+      .digest('base64')
+
+    return {
+      'X-Zendesk-Webhook-Signature-Timestamp': timestamp,
+      'X-Zendesk-Webhook-Signature': signature
+    }
+  }
+  test('API Gateway returns a 404 response if the request refers to a non-existent Zendesk ticket', async () => {
```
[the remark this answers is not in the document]
gdsandrewmccafferty: Yeah, I don't think there's really any need to check the logs for something if the response comes back with this precise information. The "check the logs" code is IMO something we do just because we can't figure out what's going on from the response alone.

integration-tests/zendeskRequestAPI.spec.ts
gdsandrewmccafferty: I agree with @rikumolugds . The code to create a ticket is already there in the codebase, so could you try just adding it and seeing if it works? Shouldn't be too big a job, and I agree that hard-coding something that could be removed at any point seems like a bit of a bad idea.

gdsandrewmccafferty (dismissed):

gdsandrewmccafferty (approved):

PR #71, opened 2022-09-29
Feature/tt2 57 integration tests

rikumolugds:
The first suite of integration tests comprising tests for : 
- validation of zendesk requests by the first lambda triggered by the webhook
- checking the integrity of zendesk requests
Helper functions for future zendesk and AWS cloudwatch calls have been added.  Note that environment variables need to be set as stated in the README to run the tests locally.


integration-tests/validatePIIRequest.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    await approveZendeskRequest(ticketID)
+
+    // Cloudwatch - Fetch latest log stream containing the ticket details
+    let latestLogStreamName = await getLatestLogStreamName()
+    console.log(`LATEST LOG STREAM NAME: ${latestLogStreamName}`)
+    let logStreamRequestID
+    let eventMatched = false
+
+    while (!eventMatched) {
```
SamBerryKC, line 25: Is it worth breaking out waiting on a matching event into a separate function which returns a promise we can await on?

integration-tests/lib/requestData.ts
```diff
@@ -0,0 +1,87 @@
...
+  generateZendeskRequestDate
+} from '../utils/helpers'
+
+const validRequestData = {
+  request: {
+    subject: `Integration Test Request - ` + generateRandomNumber(),
+    ticket_form_id: 5603412248860,
+    custom_fields: [
+      {
+        id: 5605352623260,
```
gdsandrewmccafferty, line 12: At least adding some constants for all these "magic numbers" would make it clearer what they're all for.
That way, if they ever change, it'll be a lot easier to see what they're for. Just name the variables with something like the name they're labelled with in Zendesk.

integration-tests/lib/zendeskParameters.ts
```diff
@@ -0,0 +1,40 @@
+const message = ' is not set. Please check your test environment file.'
+
+const getEndUsername = () => {
+  if (process.env.ZENDESK_END_USERNAME) {
+    return process.env.ZENDESK_END_USERNAME
```
gdsandrewmccafferty, line 5: You can avoid repeating yourself in these methods by using the fact that you can do process.env['ENV_VARIABLE_NAME']. Then you can write your logic for checking if the variable is set once in one function, and use it in all the others, passing in a string representing the variable name (e.g. 'ZENDESK_END_USERNAME')

integration-tests/validatePIIRequest.spec.ts
gdsandrewmccafferty: Yes, I would agree, especially after I notice that you're doing exactly the same thing twice in both tests, the code looks to be duplicated, so should be put out into another function for that reason too.

integration-tests/validatePIIRequest.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+        )
+      })
+      if (matchingEvent.length == 1) {
+        console.log(
+          `Event for ticket ${ticketID} found in log stream: ${latestLogStreamName}`
+        )
+        const message: string = matchingEvent[0].message as string
+        console.log(`MATCHED EVENT: ${message}`)
+        logStreamRequestID = extractRequestID(message)
+        eventMatched = true
```
gdsandrewmccafferty, line 47: I've not run the code, so not 100% sure, but couldn't you just break here?

integration-tests/validatePIIRequest.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+      } else {
+        latestLogStreamName = await getLatestLogStreamName()
+      }
+    }
+
+    // filter for ticket's successful validation event in the same log stream
+    const validRequestFilterPattern = `"${logStreamRequestID}" transfer`
+    console.log(`VALIDATION FILTER PATTERN: ${validRequestFilterPattern}`)
+    let validationEvents: FilteredLogEvent[] = []
```
SamBerry-GDS, line 56: This can be a const which equals the output of the function below

integration-tests/validatePIIRequest.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    validationEvents = await getMatchingLogEvents(
+      validRequestFilterPattern,
+      latestLogStreamName
+    )
+
+    expect(validationEvents.length).toEqual(1)
+    console.log(`VALIDATION EVENT: ${validationEvents[0].message}`)
```
SamBerry-GDS, line 63: Is it worth validating the message as well as that a validationEvent exists?

integration-tests/lib/requestData.ts
rikumolugds: Done.

integration-tests/validatePIIRequest.spec.ts
rikumolugds: The loop would break there as there isn't any more code to run after that. I've added `break` for more clarity.

integration-tests/validatePIIRequest.spec.ts
rikumolugds: Done

integration-tests/validatePIIRequest.spec.ts
rikumolugds: Done

integration-tests/validatePIIRequest.spec.ts
rikumolugds: Good point. Done.

integration-tests/lib/zendeskParameters.ts
rikumolugds: @gdsandrewmccafferty thanks for that tip. Done.

gdsandrewmccafferty (approved):

integration-tests/utils/helpers.ts
```diff
@@ -0,0 +1,173 @@
...
+): Promise<EventLogStream> => {
+  if (eventMessagePatterns.length == 0) {
+    throw Error('No message patterns provided for event')
+  }
+
+  let latestLogStreamName = await getLatestLogStreamName()
+  console.log(`LATEST LOG STREAM NAME: ${latestLogStreamName}`)
+  let eventMatched = false
+  // eslint-disable-next-line @typescript-eslint/no-inferrable-types
+  let eventMessage: string = ''
```
SamBerry-GDS, line 82: You can drop the type and remove the linter disable comments on these as typescript will infer the type and then you can't easily change it.

rikumolugds: I had done that but the code wouldn't compile due to `possibly unassigned` error.  I had to assign it to avoid the error. I couldn't think of any other way to clear the error.

SamBerry-GDS (approved):

PR #73, opened 2022-09-30
FEATURE/TT2-17 Generate sql for Athena query

gds-topickering:
As discussed, the current implementation accounts for Zendesk requests which feature specified dataPaths, and does not consider the looser piiTypes option


src/services/athena/createQuerySql.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const sqlSelectStatement = formatSelectStatement(requestData.dataPaths)
+
+  const sqlWhereStatement = formatWhereStatment(identifierType, identifiers)
+
+  const dataSource = `${getEnv('ATHENA_DATABASE_NAME')}.${getEnv(
+    'ATHENA_TABLE_NAME'
+  )}`
+
+  const queryString = `SELECT ${sqlSelectStatement} FROM ${dataSource} WHERE ${sqlWhereStatement}`
```
gdsandrewmccafferty, line 37: Is it possible to do parameterised queries in Athena? Just thinking, this looks it could be a bit Bobby Tables https://xkcd.com/327/?

gds-topickering: Yes, this is a good point and one I had some concern about. At present Athena only really allows parameterisation of the 'WHERE' parts, but there may be another way using what are termed Prepared statements - I'll investigate further.

gds-topickering: Prepared Statements have the same limitation (i.e. only allowing parameterisation of the WHERE clause). But this is something that should be addressed as it is a risk.

gds-topickering: Athena only supports parameterisation of the WHERE clause -  I have updated the  function to return a parameterised query to the extent supported.  The other values are all validated at various points - piiTypes and indentifierTypes are defined types and the dataPaths are validated for format.

src/lambdas/initiateAthenaQuery/handler.spec.ts
```diff
@@ -14,26 +15,37 @@ jest.mock('../../sharedServices/zendesk/updateZendeskTicket', () => ({
...
   Promise<ConfirmAthenaTableResult>
 >
 const mockUpdateZendeskTicket = updateZendeskTicketById as jest.Mock
 const mockGetQueryByZendeskId = getQueryByZendeskId as jest.Mock
+const mockCreateQuerySql = createQuerySql as jest.Mock
 
 describe('initiate athena query handler', () => {
   beforeEach(() => {
     mockConfirmAthenaTable.mockReset()
+    mockCreateQuerySql.mockReset()
```
gdsandrewmccafferty, line 32: You can reset all your mocks in one go with `jest.resetMocks()`, a tip I got from Mabon!

src/lambdas/initiateAthenaQuery/handler.spec.ts
```diff
@@ -55,4 +67,23 @@ describe('initiate athena query handler', () => {
...
+    mockCreateQuerySql.mockReturnValue({
+      sqlGenerated: false,
+      error: 'sql error message'
+    })
+
+    await expect(handler(testAthenaQueryEvent)).rejects.toThrow(
+      'sql error message'
+    )
+    expect(mockGetQueryByZendeskId).toHaveBeenCalled()
+    expect(mockConfirmAthenaTable).toHaveBeenCalled()
```
gdsandrewmccafferty, line 85: Can you use `toHaveBeenCalledWith` here, and assert on the parameters you're expecting?

src/sharedServices/athena/createQuerySql.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { createQuerySql } from './createQuerySql'
+import {
+  testDataRequest,
+  noEventIdTestDataRequest,
+  dataPathsTestDataRequest
+} from '../../utils/tests/testDataRequest'
+
+describe('create Query SQL', () => {
+  test('returns a formatted SQL query string if all conditions satisfied', () => {
+    expect(createQuerySql(dataPathsTestDataRequest)).toEqual({
```
gdsandrewmccafferty, line 10: Because you have different types of identifier, I'd like to see a parameterised test to try these different scenarios out. There's a Jest thing called `it.each`

src/sharedServices/athena/createQuerySql.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+      sqlGenerated: true,
+      sql: "SELECT json_extract(restricted, '$.user.firstName') as user_firstname, json_extract(restricted, '$.user.lastName') as user_lastname FROM test_database.test_table WHERE event_id=? OR event_id=?",
```
gdsandrewmccafferty, line 12: I think you can use the `IN` operator here, at least according to my googling

https://docs.aws.amazon.com/athena/latest/ug/select.html

```
SELECT * FROM "webdata"."impressions"
WHERE referrer IN ('example.com','example.net','example.org')
```

gds-topickering: Done

src/lambdas/initiateAthenaQuery/handler.spec.ts
gds-topickering: Thanks

src/sharedServices/athena/createQuerySql.spec.ts
gds-topickering: Done

src/lambdas/initiateAthenaQuery/handler.spec.ts
gds-topickering: Done

src/sharedServices/athena/createQuerySql.spec.ts
gdsandrewmccafferty: Better, but can you also parameterise the happy path? i.e. `returns a formatted SQL query string if all conditions satisfied` needs one for each type of id, because you specifically deal with each in the code

gdsandrewmccafferty (approved):

PR #83, opened 2022-10-05
Groups handler and service functions

SamBerry-GDS:
Handlers are moved into folders with code specific to them.

Route into each handler is now index to differentiate from other files.

Services which are shared are now grouped by service under sharedServices rather than all sitting in services.


andylangridge (changes requested):
Can we change `handlers` to `functions`? And any `index.ts` files to `handler.ts`? 🙏🏻

SamBerry-GDS:
> Can we change `handlers` to `functions`? And any `index.ts` files to `handler.ts`? 🙏🏻

I've updated handlers to lambdas and changed the entry point files from index to  handler.

andylangridge (approved):

gdsandrewmccafferty (approved):

PR #86, opened 2022-10-06
Feature/tt2 15 process new data retrieval message

gdsandrewmccafferty:


src/sharedServices/zendesk/validateZendeskRequest.ts
```diff
@@ -151,9 +151,9 @@ const getTodayUtc = (): number => {
   return Date.UTC(today.getFullYear(), today.getMonth(), today.getDate())
 }
 
-const mapSpaceSeparatedStringToList = (input: string) => {
+const mapSpaceSeparatedStringToList = (input: string): string[] => {
```
andylangridge, line 154: Don't know if Git will take care of this, but since I merged my PR I moved this function into the `helpers.ts` file, (and moved this file into the `lambdas/initiateDataRequest` folder)

gdsandrewmccafferty: Yeah, I think I've maybe got that now after a merge, I'll check

src/sharedServices/dynamoDB/dynamoDBPut.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  if (glacierRestoreInitiated) {
+    newRecord.checkGlacierStatusCount = { N: '0' }
+  }
+  if (copyFromAuditBucketInitiated) {
+    newRecord.checkCopyStatusCount = { N: '0' }
+  }
+
+  await ddbClient.send(
```
SamBerry-GDS, line 51: It might be nicer to return the promise and not have this as an async function if we're not doing anything with result.

gdsandrewmccafferty: OK, done!

template.yaml
```diff
@@ -572,6 +592,31 @@ Resources:
                 -  "s3:GetObjectVersion"
               Resource:
                 - !Join ['/', [!GetAtt BatchJobManifestBucket.Arn, '*']]
+      - PolicyName: AllowS3Copy
+        PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+            - Effect: Allow
```
andylangridge, line 599: Can we put some Sid entries on these policies so it's a bit clearer what they're for?

i.e. AllowCopyToAnalysisBucket.

Or maybe they should be separate policies?

Basically one policy is for the audit bucket, and another for the analysis bucket which is a tad confusing to me!

src/sharedServices/queue/sendSqsMessage.ts
```diff
@@ -7,20 +7,29 @@ import { getEnv } from '../../utils/helpers'
...
+    queueUrl,
+    delaySendInSeconds
+  )
 }
 
 export const sendSqsMessageWithStringBody = async (
   messageBody: string,
-  queueUrl: string
+  queueUrl: string,
+  delaySendInSeconds: number | undefined = undefined
```
SamBerry-GDS, line 23: could we just do `delaySendInSeconds?: number`

gdsandrewmccafferty: Yes, good point

SamBerry-GDS (approved):

template.yaml
gdsandrewmccafferty: I've added Sids to everything now, is that better? I'm reluctant to pull it out any more, unless you think there's a good reason for it?

andylangridge (approved):

andylangridge (approved):

rikumolugds (approved):

PR #90, opened 2022-10-10
Feature/tt2 18 execute athena query

gds-topickering:
Adds services to:
- initiate the execution of an athena query using SQL generated based on zendesk request
- update the dynamoDb table with the athenaQueryId if query successfully initiated


src/sharedServices/dynamoDB/dynamoDBUpdate.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const updateQueryByZendeskId = async (
+  zendeskId: string,
+  attributeKey: string,
+  attributeValue: string
+): Promise<DynamoDBParams> => {
+  const params: UpdateItemCommandInput = {
+    TableName: getEnv('DYNAMODB_TABLE_NAME'),
+    Key: { zendeskId: { S: zendeskId } },
+    ReturnValues: 'ALL_NEW',
```
gds-topickering, line 18: NOTE - at present the update call returns all of the entry in the DB, including the updated fields. It is possible to configure it so that only the updated attributes are returned.

src/sharedServices/dynamoDB/dynamoDBUpdate.ts
```diff
@@ -0,0 +1,59 @@
...
+    ExpressionAttributeValues: { ':value': { S: `${attributeValue}` } },
+    UpdateExpression: `SET ${attributeKey}=:value`
+  }
+
+  const updatedData = await ddbClient.send(new UpdateItemCommand(params))
+  const responseObject = updatedData?.Attributes
+
+  if (!responseObject) {
+    throw new Error(
```
andylangridge, line 27: Do we need to do anything here? Like close the Zendesk ticket?

Same question for the `throw` further down too!

gds-topickering: I've added a means to handle errors and update the zendesk ticket in the handler

src/sharedServices/dynamoDB/dynamoDBUpdate.ts
andylangridge: I think it's fine to return the whole entry 👍🏻

src/types/dynamoDBParams.ts
```diff
@@ -0,0 +1,5 @@
+import { DataRequestParams } from './dataRequestParams'
+
+export interface DynamoDBParams extends DataRequestParams {
```
andylangridge, line 3: Does the table we use have a more descriptive name? `DynamoDBParams` feels a bit generic to me and could be anything. Especially considering we're introducing another table soon for the secure fraud site stuff.

gdsandrewmccafferty: As above, we won't need this interface at all once we stop deserializing stuff

src/lambdas/initiateAthenaQuery/startQueryExecution.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      QueryExecutionId: '123'
+    })
+
+    const result = await startQueryExecution({
+      sqlGenerated: true,
+      idParameters: ['test_parameter'],
+      sql: 'test sql'
+    })
+    expect(result).toEqual({
+      queryExecuted: true,
```
gdsandrewmccafferty, line 26: You can use the `expect(athenaMock).toHaveReceivedCommandWith` with the relevant Command to assert that the right parameters were passed in. For an example, see `src/sharedServices/queue/sendSqsMessage.spec.ts`

src/sharedServices/dynamoDB/dynamoDBUpdate.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    dynamoMock
+      .on(UpdateItemCommand)
+      .resolves(mockUpdatedDbContents as UpdateItemOutput)
+
+    const result = await updateQueryByZendeskId(
+      '12',
+      'athenaQueryId',
+      'testAthenaId'
+    )
+    expect(result).toEqual({
```
gdsandrewmccafferty, line 47: Similarly to above, you can check that the right commands were passed using `toHaveReceivedCommandWith`

src/types/dynamoDBParams.ts
gds-topickering: I've kept the interface but used it in the dynamoDBGet service - this will be used elsewhere so  needed to be expanded to account for other fields such as the athenaQueryId

src/sharedServices/dynamoDB/dynamoDBUpdate.spec.ts
gds-topickering: Done

src/lambdas/initiateAthenaQuery/handler.ts
```diff
@@ -1,46 +1,114 @@
...
+  await confirmQuerySqlGeneration(querySql, zendeskId)
+
+  const queryExecutionDetails = await startQueryExecution(querySql)
+
+  await confirmQueryExecution(queryExecutionDetails, zendeskId)
+
+  return
+}
+
+const confirmEvent = (event: SQSEvent): string => {
```
gdsandrewmccafferty, line 34: Minor thing, but you could just call this `retrieveZendeskIdFromEvent` for clarity

gds-topickering: Done

gdsandrewmccafferty (dismissed):
Just resolve the merge conflicts, then I'm happy with this now!

gdsandrewmccafferty (approved):

PR #91, opened 2022-10-10
Feature/TT2-71 - Interpolate and move strings

kasgds:
This updates includes a template function to generate strings.


andylangridge (changes requested):
Can we call the folder `constants` or (something else), instead of `i18n`? Couple of reasons for this:
1. I don't think we would every be localising log copy
2. In my opinion it's confusing to have things we _might_ need in future (YAGNI principal and all that)

kasgds:
> Can we call the folder `constants` or (something else), instead of `i18n`? Couple of reasons for this:
> 
> 1. I don't think we would every be localising log copy
> 2. In my opinion it's confusing to have things we _might_ need in future (YAGNI principal and all that)

`constants` makes sense so I've renamed it to that.

src/utils/interpolateTemplate.ts
```diff
@@ -0,0 +1,40 @@
+interface MessagesInterface {
```
gdsandrewmccafferty, line 1: I'd like to see some tests for the `interpolateTemplate` method, given that you've got a fair amount of logic in there, and everything else in the tests depend on this working properly.

Alternatively, you could put all the tests back to working how they did before (i.e. just assert on all the strings as before), and then you've got validation that we've not regressed.

andylangridge (dismissed):

src/utils/interpolateTemplate.ts
kasgds: Fair point. I've undone the changes to the spec files.

gdsandrewmccafferty (approved):
Yep, better now that we've got the tests checking that the functionality works

PR #92, opened 2022-10-11
Fix integration tests by adding new required fields

gdsandrewmccafferty:


integration-tests/lib/requestData.ts
```diff
@@ -10,6 +11,8 @@ const PII_FORM_REQUEST_DATE_FIELD_ID = 5605700069916
 const PII_FORM_REQUESTED_PII_TYPE_FIELD_ID = 5641719421852
 const PII_FORM_CUSTOM_DATA_PATH_FIELD_ID = 5698447116060
 const ZENDESK_SUPPORT_PII_REQUEST_STATUS_FIELD_ID = 5605885870748
+const PII_FORM_IDENTIFIER_RECIPIENT_EMAIL = 6202354485660
+const PII_FORM_IDENTIFIER_RECEIPIENT_NAME = 6202301182364
 
 const validRequestData = {
```
rikumolugds, line 17: With these new fields, there is now a second case of valid data, when the recipient's name/email is different from the requester's. We can make up a random email, or I can add that in my next PR.

gdsandrewmccafferty: I would add that in another PR, because this one is really just to get things working again

andylangridge: I'd also add that scenario is covered by unit tests. Integration tests (in my opinion) should really only cover the journeys. Since they take a lot longer to run, there is no need to cover cases like this. Else we get ourselves into integration test hell with a suite that takes hours to finish 😅

andylangridge (approved):

rikumolugds (approved):

PR #93, opened 2022-10-11
Add workflow_dispatch to relevant conditions

andylangridge:
Manually running deployments was not working as every step was being skipped. This includes the event `workflow_dispatch` (which represents manual deployments) in the conditions for the relevant jobs.


gdsandrewmccafferty (approved):
Looks fine to me, as long as it all works!

PR #97, opened 2022-10-13
Refactor structure of integation-tests folder

andylangridge:
* Added some catch statements to the Zendesk API calls to help debugging when things go wrong
* Added some types
* Move the utils functions/files into sub-folders
* Added most things as constants
* Renamed the test files as I think the old `validatePII..` test is going to be more around checking if the process starts/doesn't start, rather than various valid/invalid scenarios as we can use the API test for that
* Standardised exports - rather than having them at the bottom of the file, have moved an export word in front of each function that is exported. Personal preference from me really, because it's easier to see which functions are exported and which are not. Happy to change back though if people prefer the other way?


integration-tests/utils/zendesk/approveZendeskTicket.ts
```diff
@@ -0,0 +1,25 @@
...
+export const approveZendeskTicket = async (ticketId: string) => {
+  try {
+    await axios({
+      url: `${ZENDESK_BASE_URL}${ZENDESK_TICKETS_ENDPOINT}/${ticketId}`,
+      method: 'PUT',
+      headers: {
+        Authorization: authoriseAs(ZENDESK_AGENT_EMAIL),
+        'Content-Type': 'application/json'
+      },
+      data: ticketApprovalData
```
rikumolugds, line 19: Can we have the assertions back in a `then`? Especially to check the new ticket status.

integration-tests/initiateDataRequest.spec.ts
```diff
@@ -2,12 +2,11 @@ import {
   getMatchingLogEvents,
```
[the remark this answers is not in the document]
andylangridge: I don't see why you would do this, as these are the integration test files, so it makes sense for them to live in `integration-tests`

integration-tests/utils/zendesk/createZendeskTicket.ts
```diff
@@ -0,0 +1,26 @@
...
+    const response = await axios({
+      url: `${ZENDESK_BASE_URL}${ZENDESK_REQUESTS_ENDPOINT}`,
+      method: 'POST',
+      headers: {
+        Authorization: authoriseAs(ZENDESK_END_USER_EMAIL),
+        'Content-Type': 'application/json'
+      },
+      data: requestData
+    })
+    return response.data.request.id.toString()
```
rikumolugds, line 21: Can we log the ticket id? It would help for checking in Zendesk when debugging.

integration-tests/utils/zendesk/createZendeskTicket.ts
```diff
@@ -0,0 +1,26 @@
...
+  } catch (error) {
```
rikumolugds, line 22: Can we add back the assertions in `then`?

integration-tests/utils/zendesk/approveZendeskTicket.ts
andylangridge: I removed the assertions on purpose, as I didn't see the need to assert that Zendesk response was a 201, etc. as I don't think we need to be testing the Zendesk API.

A successful response means it has done its job surely?

integration-tests/utils/zendesk/createZendeskTicket.ts
andylangridge: Done

integration-tests/utils/zendesk/approveZendeskTicket.ts
andylangridge: Actually on further reflection, it makes sense to check the ticket contains the approved tag as this is a custom thing we configure in zendesk, so have added this back.

gdsandrewmccafferty (approved):
Looks fine to me!

rikumolugds (approved):

PR #98, opened 2022-10-14
Refactor tests to ensure all events are present

andylangridge:
* Refactored the tests that wait for logs, we now wait for the final event before we check that the log we want is present.
* Refactored so we can call 1 function from the tests, and split the logic in this function out in to additional functions

### DIFF
--- a/integration-tests/initiateDataRequest.spec.ts
+++ b/integration-tests/initiateDataRequest.spec.ts
@@ -1,4 +1,4 @@
-import { getCloudWatchLogEventsByMessagePattern } from './utils/aws/cloudWatchGetLogs'
+import { getCloudWatchLogEventsGroupByMessagePattern } from './utils/aws/cloudWatchGetLogs'
 import { createZendeskTicket } from './utils/zendesk/createZendeskTicket'
 import { approveZendeskTicket } from './utils/zendesk/approveZendeskTicket'
 import { deleteZendeskTicket } from './utils/zendesk/deleteZendeskTicket'
@@ -28,7 +28,7 @@ describe('Submit a PII request with approved ticket data', () => {
     })
 
     it('Should log a success in cloud watch if Zendesk request is valid', async () => {
-      const logStreamEvents = await getCloudWatchLogEventsByMessagePattern(
+      const logStreamEvents = await getCloudWatchLogEventsGroupByMessagePattern(
         INITIATE_DATA_REQUEST_LAMBDA_LOG_GROUP,
         'received Zendesk webhook',
         'zendeskId',
@@ -55,7 +55,7 @@ describe('Submit a PII request with approved ticket data', () => {
     })
 
     it('Should log an error in cloud watch if zendesk request is not valid', async () => {
-      const logStreamEvents = await getCloudWatchLogEventsByMessagePattern(
+      const logStreamEvents = await getCloudWatchLogEventsGroupByMessagePattern(
         INITIATE_DATA_REQUEST_LAMBDA_LOG_GROUP,
         'received Zendesk webhook',
         'zendeskId',
--- a/integration-tests/initiateDataRequest.spec.ts
+++ b/integration-tests/initiateDataRequest.spec.ts
@@ -1,12 +1,21 @@
 import {
-  getMatchingLogEvents,
-  extractRequestIDFromEventMessage,
-  waitForLogStreamContainingEvent
+  // findMatchingLogEvent,
+  // extractRequestIDFromEventMessage,
+  // getCloudWatchLogEventsForRequest,
+  getCloudWatchLogEventsByMessagePattern
 } from './utils/aws/cloudWatchGetLogs'
 import { createZendeskTicket } from './utils/zendesk/createZendeskTicket'
 import { approveZendeskTicket } from './utils/zendesk/approveZendeskTicket'
 import { deleteZendeskTicket } from './utils/zendesk/deleteZendeskTicket'
 import { invalidRequestData, validRequestData } from './constants/requestData'
+import { INITIATE_DATA_REQUEST_LAMBDA_LOG_GROUP } from './constants/awsParameters'
+import { FilteredLogEvent } from '@aws-sdk/client-cloudwatch-logs'
+
+const assertLogPresent = (logEvents: FilteredLogEvent[], message: string) => {
+  const event = logEvents.find((event) => event.message?.includes(message))
+  const eventFound = event ? true : false
+  expect(eventFound).toEqual(true)
+}
 
 describe('Submit a PII request with approved ticket data', () => {
   jest.setTimeout(60000)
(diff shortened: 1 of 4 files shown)


integration-tests/initiateDataRequest.spec.ts
```diff
@@ -24,28 +43,32 @@ describe('Submit a PII request with approved ticket data', () => {
...
+          ticketId
+        )
+
+      expect(
+        isLogPresent(initiateDataRequestEvents, DATA_SENT_TO_QUEUE_MESSAGE)
+      ).toEqual(true)
+
+      const messageId = getQueueMessageId(initiateDataRequestEvents)
+
+      if (messageId) {
```
andylangridge, line 60: Question if this is a good idea or not. Checking the next lambda involves waiting a fair while as the messave goes to a queue, and that message may not be read instantly.

* Would it be enough just to check that the log message is there with the queue id? that's inferred at the moment
* We could increase the number of tries or waits when searching for the logs? This is currently 30 attempts, perhaps we could give the option to customise this value?

andylangridge: Since no one responded, I have now added an attempts parameter, which might be useful regardless. Question still stands about checking the next lambda log group though!

integration-tests/utils/aws/cloudWatchGetLogs.ts
```diff
@@ -1,141 +1,118 @@
...
-  const filterLogEvents: FilteredLogEvent[] =
-    filterLogEventsResponse.events ?? []
+    if (logEvents.length > 1) {
+      throw Error('More than 1 event matched, check filter patterns')
+    }
 
-  console.log(`FILTERED LOG EVENTS: ${filterLogEvents.length}`)
+    const message = logEvents[0]?.message as string
+    console.log(`Found event: ${message}`)
+    eventMatched = true
```
gdsandrewmccafferty, line 110: I don't think you need this variable `eventMatched` now, because you're returning underneath, and the attempts < maxAttempts condition above is what stops the loop.

integration-tests/utils/zendesk/showZendeskTicket.ts
```diff
@@ -0,0 +1,25 @@
+import axios from 'axios'
+import { authoriseAs } from './authoriseAs'
+import {
+  ZENDESK_AGENT_EMAIL,
+  ZENDESK_BASE_URL,
+  ZENDESK_TICKETS_ENDPOINT
+} from '../../constants/zendeskParameters'
+
+export const showZendeskTicket = async (ticketId: string) => {
```
gdsandrewmccafferty, line 9: Is there a reason you're not just calling this `getZendeskTicket`? Calling it "show" makes me think there's some UI or something. Pedantic point though...

andylangridge: That's what they call it the API docs! But I see your point, I prefer `getZendeskTicket` so will rename!

integration-tests/utils/aws/cloudWatchGetLogs.ts
andylangridge: Done

integration-tests/initiateDataRequest.spec.ts
SamBerryKC: We should make sure it reaches the queue to prove out the integration up to that circuit breaker.  What you have now looks good.

integration-tests/initiateDataRequest.spec.ts
```diff
@@ -1,16 +1,49 @@
...
-  extractRequestIDFromEventMessage,
-  waitForLogStreamContainingEvent
-} from './utils/aws/cloudWatchGetLogs'
+import { getCloudWatchLogEventsGroupByMessagePattern } from './utils/aws/cloudWatchGetLogs'
 import { createZendeskTicket } from './utils/zendesk/createZendeskTicket'
 import { approveZendeskTicket } from './utils/zendesk/approveZendeskTicket'
 import { deleteZendeskTicket } from './utils/zendesk/deleteZendeskTicket'
 import { invalidRequestData, validRequestData } from './constants/requestData'
+import {
+  INITIATE_DATA_REQUEST_LAMBDA_LOG_GROUP,
```
rikumolugds, line 7: We should probably use env variables for the log groups as suggested by AndyM? I've changed to that in my local branch.

integration-tests/utils/aws/cloudWatchGetLogs.ts
```diff
@@ -1,141 +1,116 @@
...
+      logStreams,
+      eventMessagePatterns
+    )
 
-  await pause(500)
-  const filterLogEventsResponse: FilterLogEventsCommandOutput =
-    await cloudWatchLogsClient.send(filterLogEventsCommand)
+    if (logEvents.length == 0) {
+      await pause(1000)
+      logStreams = await getLogStreams(logGroupName)
```
SamBerryKC, line 99: Would this work better on the line before we get logEvents in this loop?  We could then not call it outside of the loop.

integration-tests/initiateDataRequest.spec.ts
andylangridge: Since we're going to be checking stuff later down the line, checking the next lambda is useful.

integration-tests/utils/aws/cloudWatchGetLogs.ts
```diff
@@ -1,141 +1,116 @@
...
+    eventMessagePatterns,
+    maxAttempts
   )
 
-  await pause(500)
+  if (!event) return []
 
-  const describeLogStreamsResponse: DescribeLogStreamsCommandOutput =
-    await cloudWatchLogsClient.send(describeLogStreamsCommand)
+  const requestId = extractRequestIdFromEventMessage(event.message as string)
```
SamBerryKC, line 22: This might not be a string.

You could tweak the check on line 20 to make sure it's what your expecting.  Something like: 
`  if (!event || !event.message || !event.logStreamName) return []
`

Might be worth moving into a separate function.

gdsandrewmccafferty (dismissed):

integration-tests/utils/aws/cloudWatchGetLogs.ts
```diff
@@ -1,141 +1,116 @@
...
-  DescribeLogStreamsCommandOutput,
   DescribeLogStreamsCommand,
   LogStream,
-  FilterLogEventsCommandInput,
   FilterLogEventsCommand,
-  FilterLogEventsCommandOutput,
   FilteredLogEvent
 } from '@aws-sdk/client-cloudwatch-logs'
 
-export const getLogStreamPrefix = () => {
```
rikumolugds, line 14: Can you please leave the old functions there temporarily, as my local branch needs them? I'll refactor to use them if possible, after I merge. The athena sql events are a bit different from the initiate ones.

integration-tests/utils/aws/cloudWatchGetLogs.ts
```diff
@@ -1,141 +1,116 @@
...
+    logGroupName: logGroupName,
+    orderBy: 'LastEventTime',
+    descending: true,
+    limit: 5
   }
+  const command = new DescribeLogStreamsCommand(input)
+  const response = await cloudWatchLogsClient.send(command)
+  const logStreams = response?.logStreams ?? []
 
-  let latestLogStreamName = await getLatestLogStreamName()
```
rikumolugds, line 59: Is it okay to leave the old functions temporarily? I'll refactor to using them after I merge to my local. The athena events are a bit different to the initiate and process ones, and I'm not sure if the new ones would need more refactoring to get them working for those logs.

integration-tests/utils/aws/cloudWatchGetLogs.ts
andylangridge: Done

integration-tests/utils/aws/cloudWatchGetLogs.ts
andylangridge: Done, now `if (!event || !event.message) return []` since `logStreamName` is always there (have removed the unnecessary cast)

integration-tests/utils/aws/cloudWatchGetLogs.ts
andylangridge: Can we just pair to see why you would use this? I'm happy to help get things working. Plus if you really need it then just keep it in your branch.

PR #99, opened 2022-10-17
Chore/add ticket create script

gdsandrewmccafferty:
This PR adds the ability to create a Zendesk ticket from the command line.

To avoid messiness, this branch comes off chore/fix-test-flakiness, so we'll merge that one before this one


integration-tests/constants/requestData.ts
```diff
@@ -17,9 +17,25 @@ import {
   PII_FORM_REQUEST_STATUS_FIELD_ID
 } from './zendeskParameters'
 
+const generateSubjectLine = () => {
+  const fixedSubjectLine = process.env.FIXED_SUBJECT_LINE
+  if (fixedSubjectLine) {
+    return fixedSubjectLine
+  }
+  return `Integration Test Request - ` + generateRandomNumber()
```
kasgds, line 25: Although fine could this be re-written as: 
``
  `Integration Test Request - ${generateRandomNumber()}`
``
Just seems weird that as a template string it's not all together.

integration-tests/utils/zendesk/createAndApproveTicket.ts
```diff
@@ -0,0 +1,23 @@
+import { makeApproveZendeskTicketRequest } from './approveZendeskTicket'
+import { createZendeskTicket } from './createZendeskTicket'
+import { validRequestData } from '../../constants/requestData'
+
+const createAndApproveTicket = async () => {
+  if (!process.env.FIXED_DATA_REQUEST_DATE) {
+    throw new Error(
+      'Must set FIXED_DATA_REQUEST_DATE when calling this utility'
```
kasgds, line 8: Should this be a bit more helpful and say it should be set in `env` variables? I'd personally find this confusing and would still have to open up the file to find the test to see what it meant.

integration-tests/utils/zendesk/createAndApproveTicket.ts
```diff
@@ -0,0 +1,23 @@
...
+    )
+  }
+  if (!process.env.FIXED_SUBJECT_LINE) {
+    throw new Error('Must set FIXED_SUBJECT_LINE when calling this utility')
```
kasgds, line 12: Same as above.

integration-tests/constants/requestData.ts
gdsandrewmccafferty: Yes, it came from another part of the codebase. I'll change it, it's silly.

integration-tests/utils/zendesk/createAndApproveTicket.ts
gdsandrewmccafferty: I've updated the Readme now. Everything you use here has to be set as a shell environment variable.
In any case, if you look at the `package.json`, it sets these variables anyway, so if you use the facility as now described in the Readme, you shouldn't have any confusion.

andylangridge (approved):

PR #101, opened 2022-10-18
Amend sql generation to account for datetime partitioning

gds-topickering:


src/lambdas/initiateAthenaQuery/createQuerySql.ts
```diff
@@ -41,7 +39,13 @@ export const createQuerySql = (
     'ATHENA_TABLE_NAME'
   )}`
 
-  const queryString = `SELECT ${sqlSelectStatement} FROM ${dataSource} WHERE ${sqlWhereStatement}`
+  const queryString = `SELECT ${sqlSelectStatement} FROM ${dataSource} WHERE ${sqlWhereStatement} AND datetime >= ? AND datetime <= ?`
+
+  const formattedDateFrom = formatDateFrom(requestData.dateFrom)
+  const formattedDateTo = formatDateTo(requestData.dateTo)
+
+  identifiers.push(formattedDateFrom)
```
SamBerry-GDS, line 47: could we give this a name like parameters/athenaParams now it encompasses more things?

Is it worth generating all of them in a single function so we can look in one place and easily see what they are and the sequence they are added.

src/lambdas/initiateAthenaQuery/createQuerySql.ts
```diff
@@ -105,3 +109,17 @@ const formatWhereStatment = (
 
   return `${identifierType} IN (${whereStatementsArray.join(', ')})`
 }
+
+const formatDateFrom = (dateFrom: string): string => {
+  const splitDateFrom = dateFrom.split('-')
+  splitDateFrom.push('00')
+  const formattedDateFrom = splitDateFrom.join('/')
```
kasgds, line 116: You could just return the new join date instead of storing it in a const and returning it
` return splitDateFrom.join('/')`

gds-topickering: Thanks, will do

src/lambdas/initiateAthenaQuery/createQuerySql.ts
```diff
@@ -105,3 +109,17 @@ const formatWhereStatment = (
...
+  return formattedDateFrom
+}
+
+const formatDateTo = (dateTo: string): string => {
+  const splitDateTo = dateTo.split('-')
+  splitDateTo.push('23')
+  const formattedDateTo = splitDateTo.join('/')
```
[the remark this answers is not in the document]
gds-topickering: Ditto

src/lambdas/initiateAthenaQuery/createQuerySql.ts
gds-topickering: I will make these changes - will use queryParameters and create a single function

PR #103, opened 2022-10-19
Feature/tt2 72 confirm athena query status

gdsandrewmccafferty:


src/lambdas/sendQueryResultsNotification/handler.ts
```diff
@@ -0,0 +1,67 @@
...
+  queryDetails: AthenaEBEventDetails,
+  zendeskId: string
+): Promise<void> => {
+  const queryState = queryDetails.currentState
+  switch (queryState) {
+    case 'SUCCEEDED': {
+      return
+    }
+
+    case 'QUEUED':
```
andylangridge, line 50: For my benefit, can you explain why some cases are empty? Assume that means don't do anything... But what's the difference with the 'SUCCEEDED' case? Does the `return` make the handler stop at that point too?

template.yaml
```diff
@@ -877,16 +909,16 @@ Resources:
...
-        - IndexName: athenaQueryIdIndex
-          KeySchema:
-            - AttributeName: "athenaQueryId"
-              KeyType: "HASH"
-          Projection:
-            ProjectionType: "KEYS_ONLY"
-          ProvisionedThroughput:
-            ReadCapacityUnits: 5
-            WriteCapacityUnits: 5
+      # GlobalSecondaryIndexes:
```
andylangridge, line 912: Did you mean to delete instead of commenting out?

gds-topickering: Hmm, the hope was that this would get around the inability to update a GSI in a stack but the deployment has failed, so another approach may be required.

src/lambdas/sendQueryResultsNotification/handler.ts
gds-topickering: With the SUCCEEDED case, the return doesn't stop the handler but allows it to carry on.  Any case other than SUCCEEDED will result in an error and the handler ceasing to execute at that point

andylangridge: Oh so it just uses the default and throws an error? Then do we we need these empty entries at all?

gdsandrewmccafferty: @mabon-gds , I think @andylangridge might be right, we could just let the thing fall through to the default instead of specifically calling out QUEUED, and RUNNING, couldn't we? Just put `Function was called with unexpected state: ${queryState}. Ensure the template is configured correctly` in the default? We do need the succeeded case, because otherwise it'll error when the query succeeds, which is obviously no good.

src/lambdas/sendQueryResultsNotification/queueSendResultsReadyEmail.ts
```diff
@@ -0,0 +1,21 @@
...
+  downloadHash: string
+  zendeskTicketId: string
+  recipientEmail: string
+  recipientName: string
+}) => {
+  const emailOptions: PersonalisationOptions = {
+    email: parameters.recipientEmail,
+    firstName: parameters.recipientName,
+    zendeskId: parameters.zendeskTicketId,
+    signedUrl: createSecureDownloadLink(parameters.downloadHash)
```
SamBerry-GDS, line 15: do you think we should rename this to secure download link (in notify as well)?

gdsandrewmccafferty: Yes, it is a bit misleading. Probably the best time to do it, while we still can!

gdsandrewmccafferty: This has been dealt with now

template.yaml
gdsandrewmccafferty: We'll be dealing with this during the merge to main, may be a case of trial and error to some extent

src/lambdas/sendQueryResultsNotification/handler.ts
gdsandrewmccafferty: Right, this is done now

andylangridge (approved):

PR #104, opened 2022-10-19
Feature/tt2 17 integration tests

rikumolugds:
- New tests for Athena query lambda (happy and unhappy paths)
- Helper functions for dynamodb and SQS operations in tests
- Updated valid request data object with newly required data path values and new event id format
- Refactored Zendesk form field id's into an enum for easier reference
- Some refactoring to enhance code reuse


integration-tests/constants/zendeskFormFieldIDs.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export enum ZendeskFormFieldIDs {
+  ZENDESK_PII_FORM_ID = 5603412248860,
+  PII_FORM_IDENTIFIER_FIELD_ID = 5605352623260,
+  PII_FORM_EVENT_ID_LIST_FIELD_ID = 5605423021084,
+  PII_FORM_REQUEST_DATE_FIELD_ID = 5605700069916,
+  PII_FORM_REQUESTED_PII_TYPE_FIELD_ID = 5641719421852,
+  PII_FORM_CUSTOM_DATA_PATH_FIELD_ID = 5698447116060,
+  PII_FORM_IDENTIFIER_RECIPIENT_EMAIL = 6202354485660,
+  PII_FORM_IDENTIFIER_RECEIPIENT_NAME = 6202301182364,
```
gds-topickering, line 11: This could be PII_FORM_IDENTIFIER_RECIPIENT_NAME

integration-tests/athenaQuery.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  INITIATE_ATHENA_QUERY_LAMBDA_LOG_GROUP,
+  INITIATE_ATHENA_QUERY_QUEUE_URL
+} from './constants/awsParameters'
+import { deleteZendeskTicket } from './utils/zendesk/deleteZendeskTicket'
+
+describe('Athena Query SQL generation and execution', () => {
+  jest.setTimeout(60000)
+  let ticketId: string
+
+  beforeEach(async () => {
```
andylangridge, line 23: If we're just putting messages directly in the queue. Do we need to create a Zendesk ticket? This will cause 2 conflicting processes to run surely?

integration-tests/constants/webookRequestParameters.ts
```diff
@@ -0,0 +1,22 @@
+export interface DataRequestParams {
```
andylangridge, line 1: This should live in the `types` folder, and I'd rename the file to have the same name as the main type i.e. `dataRequestParams.ts` to avoid confusion.

integration-tests/utils/aws/cloudWatchGetLogs.ts
```diff
@@ -112,3 +113,11 @@ const waitForEventWithPatterns = async (
 const pause = (delay: number): Promise<unknown> => {
   return new Promise((r) => setTimeout(r, delay))
 }
+
+export const isLogPresent = (
```
andylangridge, line 117: If you merge the latest `main` branch this function was renamed and the implementation changed slightly.

integration-tests/utils/pause.ts
```diff
@@ -0,0 +1,3 @@
+export const pause = (delay: number): Promise<unknown> => {
```
andylangridge, line 1: I think this could actually live in the `helpers.ts` file instead of its own file.

integration-tests/utils/awsClients.ts
```diff
@@ -0,0 +1,16 @@
+import { CloudWatchLogsClient } from '@aws-sdk/client-cloudwatch-logs'
+import { DynamoDBClient } from '@aws-sdk/client-dynamodb'
+import { SQSClient } from '@aws-sdk/client-sqs'
+
+const region = 'eu-west-2'
```
andylangridge, line 5: We don't need to set this variable as there is a constant in the `awsParameters.ts` file.

integration-tests/utils/awsClients.ts
```diff
@@ -0,0 +1,16 @@
...
+const cloudWatchLogsClient = new CloudWatchLogsClient({
+  region: `${region}`
```
andylangridge, line 7: You don't need to add string interpolation here. You can simply do `region: region`. Or `region: AWS_REGION` if you update to use the contant.

integration-tests/zendeskWebhookAPI.spec.ts
```diff
@@ -22,12 +22,14 @@ const defaultWebhookRequestData: ZendeskWebhookRequest = {
...
   identifierType: 'event_id',
-  eventIds: '637783 3256',
+  eventIds:
+    'c9e2bf44-b95e-4f9a-81c4-cf02d42c1555 c9e2bf44-b95e-4f9a-81c4-cf02d42cabcd',
   piiTypes: 'drivers_license',
   sessionIds: '',
   journeyIds: '',
   userIds: '',
-  dataPaths: ''
+  dataPaths:
```
andylangridge, line 31: dataPaths aren't supported yet so I don't think we need to add this?

andylangridge: Although if it won't cause any failures I guess we can just leave it there until they are supported.

rikumolugds: Yes. I added that for consistency, as I updated the others to support the current requirement of the athena lambda.

integration-tests/constants/zendeskFormFieldIDs.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable no-unused-vars */
+
+export enum ZendeskFormFieldIDs {
```
andylangridge, line 3: What are we gaining from making this an enum in a separate file? Can we move this back to the previous file?

If this was done for organisation and clarity, we can make the existing `zendeskParameters.ts` file a bit clearer with simple key value objects.

For example:
```
export const PII_FORM_FIELD_IDS = {
  IDENTIFIER: 5605352623260
  EVENT_ID: 5605423021084
  DATE: 5605700069916
  REQUESTED_PII_TYPE: 5641719421852
  CUSTOM_DATA_PATH: 5698447116060
  REQUEST_STATUS_FIELD_ID: 5605885870748
  RECIPIENT_EMAIL: 6202354485660
  RECEIPIENT_NAME: 6202301182364
}
```

Note: I left out `ZENDESK_PII_FORM_ID = 5603412248860` because this is not a field id, it's a form id.

You could also do this for the other grouping.

andylangridge: You would then reference them the same way e.g. `PII_FORM_FIELD_IDS.IDENTIFIER`

integration-tests/utils/awsClients.ts
rikumolugds: I've removed that file completely and separated the clients.

integration-tests/constants/zendeskFormFieldIDs.ts
rikumolugds: I did that for better organisation. I didn't know enums could live in the same file as other things. Done.

integration-tests/constants/dynamoDBTicketDetails.ts
```diff
@@ -0,0 +1,302 @@
...
+    due_at: null,
+    tags: [
+      'approved',
+      'drivers_license',
+      'event_id',
+      'pii_ticf_analyst',
+      'ticf_analyst'
+    ],
+    custom_fields: [
+      { id: 360021265420, value: null },
```
gdsandrewmccafferty, line 39: I presume the test would work if you left all these null fields out? If so, I would do that, because it's just wasted space and makes things harder to read.

gdsandrewmccafferty (approved):

integration-tests/constants/zendeskFormFieldIDs.ts
andylangridge: Can we not make this an enum though? An object with key values would be much better. Means we don't need to disable linting rules which you seem to have done.

integration-tests/utils/aws/dynamoDB.ts
```diff
@@ -0,0 +1,187 @@
...
+              ticketDetails,
+              ZendeskFormFieldIDs.PII_FORM_REQUESTED_PII_TYPE_FIELD_ID
+            )
+          }
+        }
+      }
+    }
+  }
+
+  let data = null
```
andylangridge, line 104: You can just declare a const in the `try` block since this is never reassigned. You can also move the `expect` inside too.

andylangridge:
I get these linting errors when I run the linting tool:

```
~di-txma-ticf-integration/integration-tests/athenaQuery.spec.ts
  59:14  warning  Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion

~/di-txma-ticf-integration/integration-tests/utils/aws/dynamoDB.ts
  114:54  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  129:49  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  170:10  warning  Forbidden non-null assertion              @typescript-eslint/no-non-null-assertion

~/di-txma-ticf-integration/integration-tests/utils/aws/sqs.ts
  4:50  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
```

Can we fix these before merging please? Preferably without ignoring them.

integration-tests/utils/aws/dynamoDB.ts
```diff
@@ -0,0 +1,187 @@
...
+            )}`
+          },
+          recipientEmail: {
+            S: `${getFieldValue(
+              ticketDetails,
+              ZendeskFormFieldIDs.PII_FORM_IDENTIFIER_RECIPIENT_EMAIL
+            )}`
+          },
+          recipientName: {
+            S: `${getFieldValue(
```
andylangridge, line 52: I don't think we need these  `getFieldValue` and `getFieldListValues` functions if we're just referencing a static object?

PR #111, opened 2022-10-20
Replace Webpack with ESbuild

andylangridge:
* Replaced Webpack with ESLint as the build tool, this is a bit faster and aligns better with other teams in the program.
* Upgraded Yarn - included the release in the `.yarn` folder as well as a `.yarnrc.yml` file that will ensure people are using the correct version.
* Upgraded Typescript to the 4.8.4 (the only reason we were on 4.7.4 was because Webpack did not support it at the time)
* Updated the aws mock library to match the same version as its jest matchers (Yarn was complaining about this silently)
* Fixed some tests that began to fail after the library upgrade


rikumolugds (commented):
LGTM as all unit tests pass. But I'll leave approval to a Dev.

SamBerry-GDS (approved):

PR #116, opened 2022-10-24
Chore/tt2 88 edge case copy integration tests

HarrietwilliamsTAB:
tests added for mix data copy and no data copy


integration-tests/initiateDataRequest.spec.ts
```diff
@@ -192,6 +198,61 @@ describe('Submit a PII request with approved ticket data', () => {
...
+      await copyAuditDataFromTestDataBucket(
+        AUDIT_BUCKET_NAME,
+        `firehose/${INTEGRATION_TEST_DATE_PREFIX_MIX_DATA}/02/${TEST_FILE_NAME}`,
+        TEST_FILE_NAME
+      )
+      ticketId = await createZendeskTicket(validMixRequestData)
+      await approveZendeskTicket(ticketId)
+    })
+
+    test('request for valid data in standard and glacier tier', async () => {
```
rikumolugds, line 228: As this is a valid request scenario, I'd rename  to `valid request with data in standard and glacier tier`

integration-tests/initiateDataRequest.spec.ts
```diff
@@ -192,6 +198,61 @@ describe('Submit a PII request with approved ticket data', () => {
     })
   })
 
+  describe('valid requests for standard and glacier copy - analysis bucket empty', () => {
```
rikumolugds, line 201: Could you add a scenario for data in both tiers already in analysis bucket? Only one test case would do, as we don't have that for the mixed tier.

integration-tests/initiateDataRequest.spec.ts
```diff
@@ -9,6 +9,8 @@ import { deleteZendeskTicket } from './utils/zendesk/deleteZendeskTicket'
 import {
   invalidRequestData,
   validGlacierRequestData,
+  validMixRequestData,
+  validNoRequestData,
```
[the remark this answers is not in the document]
HarrietwilliamsTAB: what about `validRequestNoData`?

integration-tests/initiateDataRequest.spec.ts
HarrietwilliamsTAB: sure

HarrietwilliamsTAB: i think this might just be a duplicate of the one we already have as once it's in the analysis bucket, any file is in the standard tier, regardless if it was originally stored in glacier. 
Also, if we move it a file from glaicer to the analysis bucket during the test, it won't be complete for running the test.

andylangridge: We can't have both tiers in the analysis bucket. They are all converted to standard tier. So this test isn't relevant I would say!

andylangridge: Plus the tier isn't relevant, the app works by comparing paths. It doesn't consider the data tier at all for this functionality.

andylangridge (approved):

integration-tests/initiateDataRequest.spec.ts
rikumolugds: My thinking was to add data to the analysis bucket in the set up, with the audit bucket data a mix of standard and glacier. It is similar to the `valid requests - data present in analysis bucket` scenario, but this time the audit data would be in both tiers. I thought it would make for an interesting scenario to see if it the lambda tries to defrost or copy if that were the case.

PR #125, opened 2022-10-27
Feature/tt2 72 integration test

rikumolugds:
- Test to check that the results csv generated by the athena query can be downloaded and contains requested data.
- This effectively tests the full flow (hence the file name). Plan to put the other spec files into a `components` subfolder in subsequent PR.
- Includes new environment variables added to README


integration-tests/endToEndFlow.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    expect(rows.length).toEqual(1)
+    expect(rows[0].event_id).toEqual(END_TO_END_TEST_EVENT_ID)
+    expect(rows[0].address_postalcode).toEqual(EXPECTED_POSTALCODE)
+    expect(rows[0].address_validfrom).toEqual(EXPECTED_ADDRESS_VALID_FROM_DATE)
+    expect(rows[0].birthdate_value).toEqual(EXPECTED_BIRTH_DATE)
+    expect(rows[0].name).toBeDefined()
+  })
+
+  it('Query does not match data - Empty CSV file should be downloaded', async () => {
+    const zendeskId: string = await createZendeskTicket(
```
gdsandrewmccafferty, line 131: Given that we're going to change this in future so the user is informed that there was no data available for their query, I think maybe we shouldn't have this test, because we'll have to remove it later.

integration-tests/endToEndFlow.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  const downloadResultsCSVFromLink = async (
+    s3Link: string
+  ): Promise<string> => {
+    try {
+      const response = await axios({ url: s3Link, method: 'GET' })
+      return response.data
+    } catch (error) {
+      throw 'Error downloading results csv from S3 link'
```
gdsandrewmccafferty, line 59: You've thrown away the detail of what actually went wrong here, so at least do a log so we can see!

integration-tests/endToEndFlow.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    downloadHash: string
+  ): Promise<string> => {
+    try {
+      const response = await axios({
+        url: `${getEnv('QUERY_RESULTS_SECURE_DOWNLOAD_URL')}/${downloadHash}`,
+        method: 'POST'
+      })
+      return response.data
+    } catch (error) {
+      console.log(error)
```
gdsandrewmccafferty, line 34: Probably better to use console.error to make it stand out

gdsandrewmccafferty:
It looks like you've still got packages included from when you had neat-csv etc in the package.json. Can you run yarn again and commit the result? We shouldn't have stuff we're not using in our packages

integration-tests/endToEndFlow.spec.ts
rikumolugds: Is that change coming as part of MVP? Should we leave this until the change? Happy to remove, but this could be handy to help spot any regressions in the meantime.

integration-tests/endToEndFlow.spec.ts
rikumolugds: Done.

integration-tests/endToEndFlow.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      const response = await axios({ url: s3Link, method: 'GET' })
+      return response.data
+    } catch (error) {
+      throw 'Error downloading results csv from S3 link'
+    }
+  }
+
+  const waitForDownloadHash = async (zendeskId: string): Promise<string> => {
+    let downloadHash = await retrieveSecureDownloadDbRecord(zendeskId)
+    const maxAttempts = 30
```
HarrietwilliamsTAB, line 65: why did you choose 30 max attempts?

integration-tests/endToEndFlow.spec.ts
rikumolugds: Done.

integration-tests/endToEndFlow.spec.ts
rikumolugds: Well, that was how many attempts I felt would be reasonable based on the time it took for the data to appear during manual testing, and the overall timeout I think is reasonable for the test.

HarrietwilliamsTAB: educated decision, got ya !

integration-tests/endToEndFlow.spec.ts
gdsandrewmccafferty: Yeah, fair enough, just leave it then

gdsandrewmccafferty (approved):

PR #127, opened 2022-10-28
Feature/tt2 111 restrict email recipient list

mabon-gds:
initiateDataRequest lambda now checks to make sure the email recipient is on a list of email addresses stored in an S3 bucket


src/lambdas/initiateDataRequest/isEmailInValidRecipientList.ts
```diff
@@ -0,0 +1,16 @@
...
+): Promise<boolean> => {
+  console.log(
+    `Checking if recipient email ${recipientEmail} is in the pre-defined list of recipients`
+  )
+
+  const validRecipientList = await readS3DataToString(
+    getEnv('VALID_EMAIL_RECIPIENTS_BUCKET'),
+    'valid-email-recipients.txt'
+  )
+  console.log('this is the list of recipients: ', validRecipientList)
```
SamBerry-GDS, line 14: Not sure we want to log all the emails on every request.  Once we've tested this could we update this to just say whether the validRecipientList has a length greater than 0?

gdsandrewmccafferty: Yep, definitely, to be honest this was actually just debug code!

gdsandrewmccafferty: I've updated this now

src/sharedServices/s3/readS3DataToString.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { Body } = await s3Client.send(new GetObjectCommand(commandInput))
+
+  if (Body instanceof Readable) {
+    return await streamToString(Body)
+  } else {
+    throw new Error('Valid recipient list not found')
+  }
+}
+
+const streamToString = async (stream: Stream): Promise<string> => {
```
SamBerry-GDS, line 25: This doesn't need to be async as we return the promise.

src/sharedServices/s3/readS3DataToString.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const s3Client = new S3Client({ region: getEnv('AWS_REGION') })
+  const commandInput = {
+    Bucket: bucketName,
+    Key: fileKey
+  }
+  const { Body } = await s3Client.send(new GetObjectCommand(commandInput))
+
+  if (Body instanceof Readable) {
+    return await streamToString(Body)
```
SamBerry-GDS, line 19: We don't need to await here as the caller will have to await anyway.

src/sharedServices/s3/readS3DataToString.ts
gdsandrewmccafferty: Yep, dealt with that now

HarrietwilliamsTAB (approved):

SamBerry-GDS (approved):

PR #131, opened 2022-11-01
Change Jest output for integration tests to JUnit format

rikumolugds:
- Work to add Allure reporting for the integration tests. Uses jest-junit to generate a junit-style results file (supported JS adapters not based on Jest, plus current GDS pipeline integration currently only supports junit).
- Package installation
- Test jest config setup


package.json
```diff
@@ -52,6 +52,7 @@
     "husky": "8.0.1",
     "jest": "28.1.3",
     "jest-extended": "3.1.0",
+    "jest-junit": "^14.0.1",
```
andylangridge, line 55: Version number needs to be fixed I believe. i.e. remove the `^` :)

andylangridge (approved):

package.json
gdsandrewmccafferty: Yes, we use fixed versions everywhere, to avoid unexpected upgrades

rikumolugds: Done.

PR #132, opened 2022-11-03
Test/tt2 140 sanitise data for integration tests

rikumolugds:
- New dates, event ids, and other parameters from created new test event file for the athena query tests.
- Extended scope of athena query tests to exercise flow till validation of csv results.
- Updated random zendesk id to use timestamps
- Corrected the structure of some suites in initiate data request test to fix resulting flakiness
Note: The full pack now runs green with the following exceptions:
- Ssome flakiness in the initiate data request test (which I am investigating). Looks to be wait related
- Pending fix for athena lambda handling empty string in data paths (using an empty array in the meantime)


integration-tests/utils/aws/retrieveSecureDownloadDbRecord.ts
```diff
@@ -1,6 +1,6 @@
 import { dynamoDBClient } from './dynamoDBClient'
 import { QueryCommand } from '@aws-sdk/client-dynamodb'
-import { getEnv } from '../helpers'
+import { getEnv, pause } from '../helpers'
 export const retrieveSecureDownloadDbRecord = async (
```
gdsandrewmccafferty, line 4: I don't think you're using this export, so just export `waitForDownloadHash`, as you have, and name the file `waitForDownloadHash`.

rikumolugds: Done.

integration-tests/utils/aws/waitForDownloadHash.ts
```diff
@@ -1,6 +1,6 @@
 import { dynamoDBClient } from './dynamoDBClient'
 import { QueryCommand } from '@aws-sdk/client-dynamodb'
-import { getEnv } from '../helpers'
+import { getEnv, pause } from '../helpers'
 export const retrieveSecureDownloadDbRecord = async (
```
gdsandrewmccafferty, line 4: Just one more thing, get rid of exports when you're not using them outside of the file. It makes things clearer when refactoring, because you then know exactly what's used where.

`export` is the equivalent of making a function `public` in Java. You don't need `retrieveSecureDownloadDbRecord` anywhere else than this file

gdsandrewmccafferty (approved):

PR #135, opened 2022-11-03
Feature/handle empty string datapath

gds-topickering:
Add filters and mapping to strip out any empty or whitespace strings in datapaths.  For a belt and braces approach this is done at two points:
- when validating the zendesk ticket in the initiateDataRequest handler, and
- when creating the Athena Query sql.


src/lambdas/initiateDataRequest/validateZendeskRequest.spec.ts
```diff
@@ -359,6 +359,21 @@ describe('validateZendeskRequest', () => {
     ])
   })
 
+  it(`should handle dataPaths containing empty strings`, async () => {
```
gdsandrewmccafferty, line 362: Could we add a completely empty string to this? It seems like what you've got here is some double spacing, which is also a good test, but I'd like to make sure we also handle the possibility of nothing going on, when we should get back an empty array.

gdsandrewmccafferty (changes requested):
Fine apart from the extra test case I mentioned in comments, from my point of view

src/lambdas/initiateDataRequest/validateZendeskRequest.spec.ts
gds-topickering: Done (it had to be a separate test rather than iterating over different dataPath values)

gdsandrewmccafferty (approved):

gdsandrewmccafferty (approved):

PR #139, opened 2022-11-07
Cleanup pipeline triggers and job conditions

andylangridge:
* The Check Yarn cache job takes approx the same time to complete as the validate/run tests job. Since they run in parallel, decided it was better to run this job on all pushes instead of just PRs. This lets us remove the PR trigger and just have push triggers and avoid duplicate jobs running when PRs are opened.
* This lets us clean up the trigger conditions on other jobs
* Removed the `build` configuration from `samconfig.toml` as this was never used
* Removed the Dev S3 bucket from the deploy feature branch job. Instead we pass a `--resolve-s3` flag, which will create a SAM managed S3 bucket for us, and use an existing one if it already exists

### DIFF
--- a/.github/workflows/build-test-package.yaml
+++ b/.github/workflows/build-test-package.yaml
@@ -1,11 +1,6 @@
 name: Build, Test and Package
 
-on:
-  push:
-    branches:
-      - main
-      - feature**
-  workflow_dispatch: {}
+on: [push, workflow_dispatch]
 
 env:
   AWS_REGION: eu-west-2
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test-package.yaml
```diff
@@ -33,7 +24,7 @@ jobs:
       - name: Assume AWS Validate role
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-region: ${{ env.REGION }}
+          aws-region: ${{ env.AWS_REGION }}
```
gdsandrewmccafferty, line 27: Presume you've already checked this works, since you're changing it i.e. the AWS_REGION is definitely set?

gdsandrewmccafferty (approved):

PR #145, opened 2022-11-09
Feature/tt2 148 query audit events

mabon-gds:


src/types/audit/auditQueryDataRequestDetails.ts
```diff
@@ -0,0 +1,16 @@
+export interface AuditQueryDataRequestDetails {
+  requesterEmail: string
+  requesterName: string
+  recipientEmail: string
+  recipientName: string
+  zendeskId: string
+  dateFrom: string
+  dateTo: string
+  identifierType: string
```
gds-topickering, line 9: Could we type this as an IdentifierTypes rather than string?

gdsandrewmccafferty: No, because it could literally be anything: we're trying to cope with any requests, including invalid ones. We have to keep this very loose.

gds-topickering (approved):

src/types/audit/auditEventDetails.ts
```diff
@@ -0,0 +1,30 @@
...
+  component_id: ComponentId
+  restricted?: {
+    requesterEmail: string
+    requesterName: string
+    recipientEmail: string
+    recipientName: string
+  }
+  extensions: {
+    error?: ErrorObject
+    ticket_details: Partial<AuditQueryDataRequestDetails>
```
andylangridge, line 28: Will we always get ticket details? What about the case where there is no ticket found?

PR #146, opened 2022-11-09
integration tests enhancing debug capabilities

HarrietwilliamsTAB:
reduced the number of logs outputted when integration tests are run to streamline useful info.


integration-tests/initiateDataRequest.spec.ts
```diff
@@ -116,7 +117,6 @@ describe('Submit a PII request with approved ticket data', () => {
       assertEventPresent(initiateDataRequestEvents, DATA_SENT_TO_QUEUE_MESSAGE)
 
       const messageId = getQueueMessageId(initiateDataRequestEvents)
-      console.log('messageId', messageId)
```
rikumolugds, line 119: I think the message id is useful for debugging in the logs as it's used to link more than one log stream.

integration-tests/initiateDataRequest.spec.ts
```diff
@@ -241,7 +254,6 @@ describe('Submit a PII request with approved ticket data', () => {
...
```
rikumolugds, line 244: Please leave the message id as per comment above.

integration-tests/initiateDataRequest.spec.ts
```diff
@@ -300,8 +318,6 @@ describe('Submit a PII request with approved ticket data', () => {
 
       const messageId = getQueueMessageId(initiateDataRequestEvents)
 
-      console.log('messageId', messageId)
```
rikumolugds, line 303: Please see similar comment above.

integration-tests/initiateDataRequest.spec.ts
```diff
@@ -342,7 +363,6 @@ describe('Submit a PII request with approved ticket data', () => {
       assertEventPresent(initiateDataRequestEvents, DATA_SENT_TO_QUEUE_MESSAGE)
 
       const messageId = getQueueMessageId(initiateDataRequestEvents)
-      console.log('messageId', messageId)
```
rikumolugds, line 345: Please see above.

integration-tests/utils/aws/cloudWatchGetLogs.ts
```diff
@@ -103,10 +101,6 @@ const waitForEventWithPatterns = async (
     if (logEvents.length > 1) {
       throw Error('More than 1 event matched, check filter patterns')
     }
-
-    const message = logEvents[0]?.message as string
-    console.log(`Found event: ${message}`)
```
rikumolugds, line 108: We'd be interested in the actual event matched for debugging purposes.

rikumolugds (commented):
Comments mainly around logging message id for debugging.

integration-tests/utils/aws/cloudWatchGetLogs.ts
HarrietwilliamsTAB: have added those back

andylangridge: I would argue the opposite. We don't need to log this event as it's massive and is one of the main causes of the noise. Might be more useful to log something generic out e.g. "Event found..." if you want to see where you are in the flow. Or somehow log out the name of the event you are searching for?

andylangridge (approved):

PR #156, opened 2022-11-14
Test/test environments for integration tests

rikumolugds:
This work contains config to make the integration tests run against the different test environments.

- Updating `jest.integration.config.ts` to use dev by default
- Scripts for `dev`, `build`, and `staging`.
- Updated README.


.gitignore
```diff
@@ -3,7 +3,12 @@ dist/
 
 # Jest
 coverage/
+
+# Environment files
 .integration.test.env
+.integration.test-dev.env
```
andylangridge, line 9: You can use a wildcard here e.g. `.integration.*.env`

integration-tests/endToEndFlow.spec.ts
```diff
@@ -69,7 +70,7 @@ describe('Query results generated', () => {
     expect(rows[0].name).toBeDefined()
   })
 
-  it('Query does not match data - Empty CSV file should be downloaded', async () => {
+  it.skip('Query does not match data - Empty CSV file should be downloaded', async () => {
```
andylangridge, line 73: Since this isn't running in a pipeline, you probably don't need to add any skips to the code for now. You can use Jest's various CLI flags to achieve the same results, and there's less risk of us forgetting to remove later.

jest.integration.config.ts
```diff
@@ -4,7 +4,7 @@ const config: Config.InitialOptions = {
   coveragePathIgnorePatterns: ['/node_modules/', '/dist/'],
   testPathIgnorePatterns: ['/src/'],
   preset: 'ts-jest',
-  setupFiles: ['<rootDir>/.integration.test.env'],
+  setupFiles: ['<rootDir>/.integration.test-dev.env'],
```
andylangridge, line 7: How does this work dynamically if we hardcode to one environment here?

andylangridge: I'm assuming you're overriding in the scripts in package.json

package.json
```diff
@@ -17,7 +17,9 @@
     "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\"",
     "test": "jest -c jest.config.ts",
     "test:cov": "jest -c jest.config.ts --coverage",
-    "test:integration": "jest -c jest.integration.config.ts",
+    "test:integration-dev": "jest --setupFiles ./.integration.test-dev.env -c jest.integration.config.ts",
```
andylangridge, line 20: Would this work?

`"test:integration":  "jest --setupFiles ${SETUP_FILE:-'./.integration.test-dev.env'} -c jest.integration.config.ts"`

That way we don't need a script per environment, and you can just change the setup file with an env var e.g. 
`SETUP_FILE='./.integration.test-staging.env' yarn test:integration`

.gitignore
rikumolugds: Done.

integration-tests/endToEndFlow.spec.ts
rikumolugds: I forgot to remove that! Thanks.

jest.integration.config.ts
rikumolugds: Yes I am.

package.json
andylangridge: Or even better:
```
"test:integration": "jest --setupFiles ./integration.test-${ENV:-build}.env -c jest.integration.config.ts"
```

Then it would just be:
```
ENV=staging yarn test:integration
```

I suggest using build as the default as this will be the most up to date main stack. The dev environment will only have feature branches, so values will change all the time.

>  Although, I'm not sure how we'd set up environment variables in an environment like staging, where I believe we would not be using Github actions.

Don't worry about this, we won't be using GitHub actions at all to run integration tests. These need to be run within an AWS pipeline, which I will sort out post MVP.

rikumolugds: There may also be the need to run against other envs from local machine. One would need to remember to change the environment variable.

andylangridge (dismissed):

jest.integration.config.ts
andylangridge, line 7: Can we default this to build?

andylangridge (approved):

PR #159, opened 2022-11-14
Test/tt2 151 add tests for other id types

rikumolugds:
- Adding tests for requests based on other id types to the end to end tests.
- Also includes changes to ensure use of all the PII types in the different requests.
- Added scripts to generate emailable allure report after running integration tests
- Update README with steps to run tests and generate reports
- Some refactoring around requestData to group them according to the test pertaining to them.

**Please note, the tests will be run once I have values of the resources needed  for the different test environments**


integration-tests/constants/dynamoDBItemDetails.ts
```diff
@@ -18,7 +20,7 @@ export const dynamoDBItemDataPathAndPIITypes: DynamoDBItem = {
       },
       {
         id: 6202354485660,
-        value: 'txma-team2-ticf-analyst-dev@test.gov.uk'
+        value: recipientEmail
```
gdsandrewmccafferty, line 23: Why not just take this from the constant you already have, why do you need another variable `recipientEmail` set to the same value?

rikumolugds: Good spot. Done.

gdsandrewmccafferty (approved):

PR #165, opened 2022-11-17
Feature/tt2 170 move close zendesk role

gdsandrewmccafferty:


template.yaml
```diff
@@ -522,35 +523,57 @@ Resources:
...
+                Effect: Allow
+                Action:
+                  - sqs:ReceiveMessage
+                  - sqs:DeleteMessage
+                  - sqs:GetQueueAttributes
+                Resource: !Sub 'arn:aws:sqs:${AWS::Region}:{{resolve:ssm:QueryResultsAccountNumber}}:txma-qr-infra-${Environment}-close-zendesk-ticket-queue'
+              - Sid: DecryptKmsKeys
+                Effect: Allow
+                Action:
+                  - kms:Decrypt
```
andylangridge, line 573: Just wondering if you also need `kms:GenerateDataKey*`, `ReEncrypt*` permissions too? Noticed you have them on the other PR, and I seem to remember there is some extra stuff required when KMS keys are used with SQS.

gdsandrewmccafferty: Good point, I just copied this from the original role, but it's a good chance to see if we really need it, will try

gdsandrewmccafferty: Have checked this now, and confirmed that we can do everything we need with reading from queues and stuff falling back to the DLQ with the permissions granted here.

andylangridge (approved):

PR #173, opened 2022-11-22
Read test variables from AWS

andylangridge:
* When running in the pipeline, the tests need to be able to read their variables from AWS. Have added a `setup` folder that runs some code to set the various variables. There are 3 ways to read the variables that are all in use:  
  * Stack outputs
  * SSM parameters
  * Secrets Manager

  I have also defined some things in the jest config file that are global and I think shouldn't need to change.

  Something to note about the Jest setup file. It seems as if this is not compiled by our TS Config. Jest seems to do its own thing, and save a single file in memory. This is the reason for the `modules.export..` syntax, and the fact we had to disable some linting rules.
* Updated the Yarn scripts to run the tests. We now have `yarn test:integration` which will run the tests and assume params are coming from AWS (note that some of these can be overriden - STACK_NAME, ZENDESK_RECIPIENT_EMAIL, and ZENDESK_WEBHOOK_SECRET_KEY).

  However, we also have `yarn test:integration:dev` which will take any values you add to a `.env` file. So the tests should still work as before if you use this script.
* Updated some things to do with the Allure Reports. It seems as if we were using a few things we don't need.

  Removed `jest-allure` and `jest-jasmine2` packages. These weren't needed since all Allure needs is a junit.xml file. This was being produced by the `jest-junit` package. This also means our test runner is back to Jest instead of Jasmine. As a more modern test runner, this might solve some of our test flakiness (Sam thinks Jasmine might not be handling the async stuff correctly).

  I also noticed the test reporting assumed you had Allure CLI installed on your machine. So have replaced this with a docker-compose file to spin up Allure. We can now just get the reports with `yarn allure`. Hopefully this should make the process much simpler!


andylangridge:
TODO: I need to update the README file with these changes. Will do that before merging!

package.json
```diff
@@ -10,23 +10,21 @@
...
+    "allure:stop": "docker-compose -f integration-tests/docker-compose.allure.yaml down",
     "build": "tsc --noEmit && ts-node ./esbuild.config.ts",
     "lint": "prettier . --check || exit 1 ; eslint . --max-warnings=0",
     "lint:fix": "prettier . --write ; eslint . --fix",
     "postinstall": "husky install",
     "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\"",
     "test": "jest -c jest.config.ts",
     "test:cov": "jest -c jest.config.ts --coverage",
-    "test:integration": "yarn test:cleanReports && jest --setupFiles ./.integration.test-${ENV:-build}.env -c jest.integration.config.ts -i",
-    "test:generateEmailReport": "yarn test:stopReportContainer && docker run -d --rm --name allure -p 5050:5050 -v ${PWD}/allure-results:/app/allure-results -v ${PWD}/allure-reports:/app/default-reports frankescobar/allure-docker-service && mv ${PWD}/allure-reports/emailable-report-allure-docker-service.html ${PWD}/allure-reports/TxMA-Team2-Test_Report-$(date +%d-%m-%Y).html",
```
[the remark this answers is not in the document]
andylangridge: This is generated via the `yarn allure` script. See the README which I've just updated!

integration-tests/jest.integration.config.ts
```diff
@@ -0,0 +1,31 @@
...
+    ZENDESK_ADMIN_EMAIL: 'txma-team2-ticf-admin-dev@test.gov.uk',
+    ZENDESK_AGENT_EMAIL: 'txma-team2-ticf-approver-dev@test.gov.uk',
+    ZENDESK_END_USER_EMAIL: 'txma-team2-ticf-analyst-dev@test.gov.uk',
+    ZENDESK_END_USER_NAME: 'Txma-team2-ticf-analyst-dev'
+  },
+  preset: 'ts-jest',
+  reporters: [
+    'default',
+    [
+      'jest-junit',
```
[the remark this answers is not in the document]
andylangridge: Reasons for JUnit:
- The pipeline requires a `junit.xml` file - which I assume gets uploaded to allure, so not sure that's true regarding the report detail. Although happy to be proved wrong!
- We've been using JUnit this whole time anyway as that was what was defined in the reporter part of the config (it would override whatever jest-allure would be doing)
- The `jest-allure` library hasn't been updated in 3 years i.e. it's not maintained, whereas `jest-junit` was updated 5 days ago
- `jest-allure` would force us to use Jasmine as the test runner. This is a bad thing as it's what Jest was forked from and handles async very differently I believe

Either way we definitely don't need both of these things, and since we've been using JUnit this whole time I suggest we keep that up!

RE output location, jest-junit puts this in the required location for report generation. Now I've updated the readme you can try this out following the instructions. Let me know if they're not clear!

integration-tests/setup/retrieveStackOutputs.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!outputs) {
+    throw new Error('Stack has no outputs')
+  }
+
+  return outputs
+}
+
+export const getOutputValue = (outputs: Output[], name: string) => {
+  return outputs.find((output) => output.OutputKey === name)
+    ?.OutputValue as string
```
SamBerry-GDS, line 27: Is it worth doing a type check here and throwing an error if we've not got what we're expecting?

andylangridge: Added some type checks!

integration-tests/jest.integration.config.ts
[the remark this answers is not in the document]
rikumolugds: I'll try this locally. I think we have been using both, as I have been generating reports from the jest-allure output folder. It was my fault really for forgetting to remove the other, but you have good points supporting sticking to jest-junit. I'll just double check that the report from jest-junit captures the same amount of detail, for the sake of GDS compliance.

package.json
```diff
@@ -10,23 +10,21 @@
...
-    "test:stopReportContainer": "docker stop allure",
-    "test:showAllureReport": "allure serve",
-    "test:cleanReports": "rm -f -r allure-results/ && rm -f -r allure-report/ && rm -f -r allure-reports",
-    "createTestTicket": "ZENDESK_RECIPIENT_EMAIL=$0 FIXED_DATA_REQUEST_DATE=$1 FIXED_SUBJECT_LINE=$2 OVERRIDE_EVENT_IDS=$3 DATA_PATHS=$4 ts-node scripts/createAndApproveZendeskTicket.ts"
+    "test:integration": "jest -c integration-tests/jest.integration.config.ts -i",
+    "test:integration:dev": "jest --setupFiles '<rootDir>/.env' -c integration-tests/jest.integration.config.ts -i",
```
rikumolugds, line 23: Should we have a script for running locally against other environments using their respective env files, instead of replacing the values in one file? There are quite a few of them. Also, minor point - do we want the env file to live under the integration-tests folder or under the project root directory.

integration-tests/run-allure.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Wait for UI to become available
+while ! curl --fail --silent --head http://localhost:5252; do
+  sleep 1
+done
+
+sleep 10
+
+# Open browser at latest report
+open http://localhost:5252/allure-docker-service-ui/projects/default
```
rikumolugds, line 16: Can we add a command to rename the generated report? Because I need to upload the report everyday onto Confluence, and it complains if files have the same name. I had this in a local branch, which is now behind yours `mv allure-reports/email*.html allure-reports/TxMA-Team2-Test_Report-$(date +%d-%m-%Y).html`. Thanks

integration-tests/jest.integration.config.ts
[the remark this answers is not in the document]
rikumolugds: Update: I've checked with the report from jest-junit. The details are the same, so we are good with this.

package.json
andylangridge: No, to run locally against other environments you just log in under that environment. The vars are then automatically picked up from AWS.

If you mean the .env file, then I would also say no as the point of this change is to move away from that. Going forwards we should have only a rare use for that. I left it in only to be used in the rare circumstances it might prove useful, such as testing against a dev stack.

scripts/createAndApproveZendeskTicket.ts
```diff
@@ -51,7 +51,7 @@ const generateTicketData = (): ZendeskRequestData => ({
 const createAndApproveTicket = async () => {
   checkEnvironmentVariablesSet([
     'FIXED_SUBJECT_LINE',
-    'ZENDESK_RECIPIENT_EMAIL',
+    'FIXED_RECIPIENT_EMAIL',
```
gdsandrewmccafferty, line 54: Leave this as `ZENDESK_RECIPIENT_EMAIL` please, that's maybe moved on or something since you made this PR.

scripts/createAndApproveZendeskTicket.ts
```diff
@@ -35,7 +35,7 @@ const generateTicketData = (): ZendeskRequestData => ({
       },
       {
         id: ZendeskFormFieldIDs.PII_FORM_IDENTIFIER_RECIPIENT_EMAIL,
-        value: getEnv('ZENDESK_RECIPIENT_EMAIL')
```
gdsandrewmccafferty, line 38: As below, leave this as `ZENDESK_RECIPIENT_EMAIL`

gdsandrewmccafferty (changes requested):
I assume you've been running these tests so know they work, so my only comment is about the ticket creation script, just want to stop it getting broken again!

gdsandrewmccafferty (approved):

SamBerry-GDS (approved):

PR #175, opened 2022-11-23
Feature/tt2 176  add lambda call to int tests

mabon-gds:


integration-tests/tests/types/environmentVar.ts
```diff
@@ -24,4 +24,5 @@ export type EnvironmentVar = {
     | 'SECURE_DOWNLOAD_DYNAMODB_TABLE'
     | 'QUERY_RESULTS_SECURE_DOWNLOAD_URL'
     | 'NOTIFY_API_KEY'
+    | 'DYNAMO_OPERATIONS_FUNCTION_NAME'
```
rikumolugds, line 27: Could you please update the README and Confluence with this?

integration-tests/tests/utils/aws/waitForDownloadHash.ts
```diff
@@ -29,6 +29,7 @@ const retrieveSecureDownloadDbRecord = async (
   return result
 }
 
+// does this get used at all now? Safe to remove if not?
 export const waitForDownloadHash = async (
```
rikumolugds, line 33: I don't think we need this function anymore, as we are now calling notify instead of waiting for the download hash.

integration-tests/tests/types/environmentVar.ts
rikumolugds: Also, maybe rename to DYNAMODB_OPERATIONS_FUNCTION_NAME

integration-tests/tests/utils/aws/dynamoDB.ts
```diff
@@ -1,121 +1,44 @@
...
   ZendeskFormFieldIDs,
   ZENDESK_END_USER_EMAIL,
   ZENDESK_END_USER_NAME
 } from '../../constants/zendeskParameters'
-import { dynamoDBClient } from './dynamoDBClient'
 import { DynamoDBItem, ItemDetails } from '../../types/dynamoDBItem'
+import { invokeDynamoOperationsLambda } from './invokeDynamoOperationsLambda'
+
+export const getValueFromDynamoDB = async (
+  ticketId: string,
```
kasgds, line 10: Is `ticketId` and `zendeskId` being used interchangeably? Perhaps `zendeskId` be preferred and then you can get rid of `ticketId` on line 16?

integration-tests/tests/utils/aws/dynamoDB.ts
```diff
@@ -130,60 +53,95 @@ function getFieldListValues(ticketDetails: ItemDetails, fieldID: number) {
...
-        zendeskId: { S: `${ticketId}` }
-      },
-      ProjectionExpression: attributeName
-    }
-  } else {
-    getAttributeValueParams = {
-      TableName: AUDIT_REQUEST_DYNAMODB,
-      Key: {
-        zendeskId: { S: `${ticketId}` }
+  return {
```
SamBerry-GDS, line 67: If we're just returning this object we could use paranthesis and drop the return statement

integration-tests/tests/utils/aws/dynamoDB.ts
mabon-gds: nice spot, I've pushed a change to keep it consistent (zendeskId) in that file

integration-tests/tests/utils/aws/dynamoDB.ts
mabon-gds: thanks! That's done

integration-tests/tests/utils/aws/waitForDownloadHash.ts
mabon-gds: thanks, deleted this

gdsandrewmccafferty (approved):

SamBerryKC (approved):

PR #193, opened 2022-12-01
Set test timeout globally

andylangridge:


rikumolugds (commented):
Just checking, is this timeout going to apply for each test case or across the entire suite?

rikumolugds (approved):

andylangridge:
> Just checking, is this timeout going to apply for each test case or across the entire suite?

Each test!

PR #202, opened 2022-12-07
Feature/tt2 202 split e2e and int tests

SamBerry-GDS:
Splits the e2e and integration tests into 2 different suites


gdsandrewmccafferty (approved):
Assuming you've run everything and it all still works, I'm happy!

SamBerry-GDS:
> Assuming you've run everything and it all still works, I'm happy!

I've run the 2 suites and I've generated an Allure report for each.

Where it might go wrong is the run-tests.sh script as we promote through the pipeline but there's not a nice way to test that prior to merge (I'm relatively confident the bash is fine).

PR #203, opened 2022-12-07
Allow tests to run in parallel with random date selection

andylangridge:
- For the data copy tests, we now find an empty location in both audit and analysis buckets by means of picking a random date.
- If the date is empty on both sides, we use this date for our tests
- Removed a lot of the Delete data actions in `beforeEach` and `afterEach` as we don't need these in other tests. We can simply let the data stay there
- Added a flag to the copy function that allows the object to be tagged with `autoTest=true`
- Added a lifecycle rule to delete these tagged objects within 1 day (the minimum)
- Note: Will need to raise a PR in the TxMA audit repo for the audit bucket life cycle rule
- Removed the `-i` flag so now tests can run in parallel


integration-tests/tests/utils/aws/s3GetAvailableTestDate.ts
```diff
@@ -0,0 +1,50 @@
...
+      }
+      const command = new ListObjectsCommand(input)
+      const response = await s3Client.send(command)
+
+      return !response.Contents || response.Contents.length === 0 ? true : false
+    })
+  )
+
+  if (locationAvailability.includes(false)) {
+    await findAvailableS3Locations(generateRandomDateAndPrefix())
```
gdsandrewmccafferty, line 26: I would feel less queasy about this if it had some kind of maximum number of retries.

integration-tests/tests/utils/aws/s3GetAvailableTestDate.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  return checkDate
+}
+
+const generateRandomDateAndPrefix = (): MatchedDateAndPrefix => {
+  const date = new Date(Math.random() * Date.now())
```
gdsandrewmccafferty, line 33: What would happen if we got a clash here? i.e. would it make the tests fail? What's the probability of it coming up with something that's the same as a recently selected date? Just wondering if there's maybe a way to make this very unlikely to pick two dates the same by using the current times to scale things rather than just randomising it out.

andylangridge: As discussed, we will just pick a new date. Adding the max number of tries will stop it retrying forever too.

tests/integration-tests/test-suites/initiateDataRequestDataCopy.spec.ts
```diff
@@ -173,27 +176,27 @@ describe('Data should be copied to analysis bucket', () => {
...
+      const response = await copyAuditDataFromTestDataBucket(
+        getEnv('AUDIT_BUCKET_NAME'),
+        `${availableDate.prefix}/01/${TEST_FILE_NAME}`,
         TEST_FILE_NAME,
-        'GLACIER'
+        'GLACIER',
+        true
       )
+
+      console.log(response)
```
SamBerry-GDS, line 190: Is it worth logging here?  We're not catching errors from copyAuditDataFromTestDataBucket so if we get here it's worked.

SamBerry-GDS (approved):

SamBerry-GDS (approved):

tests/integration-tests/test-suites/initiateDataRequestDataCopy.spec.ts
andylangridge: Fixed

PR #208, opened 2022-12-08
Feature/tt2 173  implement mock server to int tests

mabon-gds:
This PR modifies the integration tests so that they make sure of the test cases set up by the mock server https://github.com/alphagov/di-txma-ticf-integration-mock-server.

Whereas before the test started by creating a ticket in Zendesk which would then trigger our lambdas, instead now what we have to do is simply call the lambda webhook from the test, using a special zendesk ID which encodes information about the test case.

The system is that the first 8 digits represent the date for the data query, then there are 5 random digits, then two last digits to indicate the test case (so we could have up to 99 test cases, although we don't need nearly that many).

An example mock Zendesk ID looks like this

202204041234501

This would mean the date for the data is 4th April 2022, and the test case is 1.

Separately, there is also some refactoring to make it clearer, when a log event is not found, which test it failed in, by making the test methods which do this return a boolean which is then asserted on in the test.


integration-tests/reports/allure-results/junit.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
andylangridge, line 1: Need to un-commit this :)

tests/integration-tests/test-suites/initiateDataRequestNegativeTests.spec.ts
```diff
@@ -3,112 +3,81 @@ import {
...
 
-      assertEventPresent(
+      const isWebhookInvalidMessageInLogs = assertEventPresent(
         initiateDataRequestEvents,
         cloudwatchLogFilters.webhookInvalid
       )
-      assertEventNotPresent(
+      expect(isWebhookInvalidMessageInLogs).toBe(true)
+
+      const isDataSentToQueueMessageNotInLogs = assertEventNotPresent(
```
andylangridge, line 38: Do we need an `assertEventNotPresent` function now that `assertEventPresent` returns a boolean?

tests/shared-test-code/constants/zendeskParameters.ts
```diff
@@ -1,3 +1,14 @@
+export const CLOSE_ZENDESK_TICKET_COMMENT =
+  'Your ticket has been closed because some fields were invalid. Here is the list of what was wrong: From Date is in the future, To Date is in the future'
+
+export const ZENDESK_END_USER_NAME = 'Txma-team2-ticf-analyst-dev'
```
andylangridge, line 4: I think we need to remove these? Might be they are still here due to all the conflicts. The idea now is that We have an object e.g. `zendeskConstants` which has a type where all the keys are readonly. This way you get auto-complete and things are a bit more organised.

Plus, we wanted to get rid of all the upper snake case variables knocking about. Although this is fine for environment vars since they are passed in from the shell (where this would be the convention in bash).

integration-tests/reports/allure-results/junit.xml
gdsandrewmccafferty: Done a while back

HarrietwilliamsTAB (approved):

andylangridge (approved):

PR #210, opened 2022-12-09
Clean up test constants and make readonly

andylangridge:
- tidy up test constants - make casing consistent (removed upper snake case and replaced with camelCase), moved constants into more relevant locations/files
- mark constants as readonly with custom types - allows autocompletion and stops them being overridden


tests/e2e-tests/constants/testData.ts
```diff
@@ -1,10 +1,22 @@
...
-export const END_TO_END_TEST_DATE_PREFIX = '2022/04/01'
-export const END_TO_END_TEST_DATE = '2022-04-01'
-export const END_TO_END_TEST_DATA_PATH =
-  'restricted.name[0].nameParts[0].value restricted.name[0].nameParts[1].value restricted.birthDate[0].value restricted.address[0].validFrom restricted.address[1].postalCode'
-export const END_TO_END_TEST_USER_ID =
-  'urn:uuid:03d3c6d3-6d7c-41df-bf45-c94401c96a2e'
-export const END_TO_END_TEST_SESSION_ID = '25bc52a5-3506-4d7a-8129-adc13a3152bf'
-export const END_TO_END_TEST_JOURNEY_ID = '71b8300f-8f06-40e7-a53e-a194beacd33f'
-export const END_TO_END_TEST_EVENT_ID = 'b122aa00-129b-46a9-b5f7-6b1bf07d427b'
+export const endToEndTestData: EndToEndTestDataConstants = {
```
SamBerry-GDS, line 1: I think we can do away with endToEndTest as it's in the e2e suite.  Maybe this could be something like:
requestValues

tests/integration-tests/constants/testData.ts
```diff
@@ -1,6 +1,15 @@
-export const ATHENA_QUERY_TEST_FILE_NAME = 'athena-query-test-data.gz'
-export const ATHENA_QUERY_DATA_TEST_DATE_PREFIX = '2022/04/01'
-export const TEST_FILE_NAME = 'test-audit-data.gz'
-export const TEST_DATA_EVENT_ID = 'c9e2bf44-b95e-4f9a-81c4-cf02d42c1552'
-export const TEST_DATA_DATA_PATHS =
-  'restricted.this1.that1 restricted.this2.that2 restricted.this3.that3.those3'
+export const integrationTestData: IntegrationTestDataConstants = {
```
SamBerry-GDS, line 1: could this just be testData

tests/integration-tests/test-suites/athenaQuery.spec.ts
```diff
@@ -128,42 +147,42 @@ describe('Athena Query SQL generation and execution', () => {
 
       const csvRows = await downloadResultsFileAndParseData(randomTicketId)
 
+      const expectedName = `[{"nameparts":[{"type":"GivenName","value":"MICHELLE"},{"type":"FamilyName","value":"KABIR"}]}]`
```
SamBerry-GDS, line 150: Could these live on the testData const?  We define them in 2 places in this file

tests/e2e-tests/constants/testData.ts
andylangridge: What about just "testData"? This is all the stuff that lives in the file, and also has the test file name too.

SamBerry-GDS: I'm fine with that - I think they are all values we pipe in to Zendesk but you can keep it more generic if you want.

tests/integration-tests/test-suites/athenaQuery.spec.ts
andylangridge: Fixed

tests/integration-tests/constants/testData.ts
andylangridge: Fixed

tests/e2e-tests/constants/testData.ts
andylangridge: Fixed

SamBerry-GDS (approved):

PR #221, opened 2023-01-25
Feature/tt2 184 update logging to use aws lambda powertools

zaydgds:


src/lambdas/processDataRequest/handler.ts
```diff
@@ -10,15 +10,20 @@ import {
...
+  initialiseLogger(context)
+  logger.info('Handling data request SQS event', { handledEvent: event })
   if (event.Records.length === 0) {
     throw new Error('No data in event')
   }
   const eventData = tryParseJSON(event.Records[0].body)
   if (isEmpty(eventData)) {
     throw new Error('Event data did not include a valid JSON body')
   }
+  logger.appendKeys({ zendeskId: eventData.zendeskId })
```
andylangridge, line 25: Is there not a function that does this?

gdsandrewmccafferty: Yep, good point. Done

andylangridge (approved):

PR #231, opened 2023-02-08
Feature/allow multiple dates

gdsandrewmccafferty:
We have a new requirement to allow TiCF analysts to specify multiple dates instead of the single date currently allowed.

We originally built the system to accept a date range, but we found from user feedback that they were much more likely to want to specify a number of disjoint dates.

The changes here are backwards compatible, so the Zendesk webhook can continue sending the original data until we decide to switch over.

However, beyond the initiateDataRequest lambda, everything has been rewritten to accept a list of dates (with the current single date a special case of this).


src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.ts
```diff
@@ -126,12 +154,9 @@ const ticketAndRequestDetailsDiffer = (
     unmatchedParameters.push('recipientName')
   if (!matchStringParams(requesterDetails.email, requestParams.requesterEmail))
     unmatchedParameters.push('requesterEmail')
+  // TODO: add check for date matching
```
andylangridge, line 157: Does this need doing in this PR? Notice upi jabe a "datesMatch" function you're using above!

src/types/queryRequestDBParams.ts
```diff
@@ -14,8 +14,7 @@ export const isQueryRequestDBParams = (
     typeof test?.recipientName === 'string' &&
     typeof test?.requesterEmail === 'string' &&
     typeof test?.requesterName === 'string' &&
-    typeof test?.dateFrom === 'string' &&
-    typeof test?.dateTo === 'string' &&
+    // TODO: add a check for the "dates" property
```
andylangridge, line 17: Another question about whether we need to do this in this PR?

src/lambdas/initiateAthenaQuery/createQuerySql.ts
```diff
@@ -178,23 +170,13 @@ const formatIdWhereStatement = (identifierType: IdentifierTypes) => {
...
-  splitDateFrom.push('00')
-  return `'${splitDateFrom.join('/')}'`
-}
-
-const formatDateTo = (dateTo: string): string => {
-  const splitDateTo = dateTo.split('-')
-  splitDateTo.push('23')
-  return `'${splitDateTo.join('/')}'`
+const formatDateForQuery = (dateFrom: string): string => {
+  return dateFrom.replaceAll('-', '/')
```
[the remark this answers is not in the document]
gdsandrewmccafferty: Yes, talking to Tom, we realised we were unnecessarily partitioning for hours, only to then say "tell me everything on this day".

You'll notice a change to the Athena table as part of this PR, where we switch to partitioning by day instead of hour.

If at some point in the future we get the analysts to tell us a particular time of day when we're looking for an event, we can put this partitioning back, we can put it back, but for now it just wasn't adding anything, and would have made the query unnecessarily complicated when we added multiple dates.

src/types/queryRequestDBParams.ts
gdsandrewmccafferty: Actually, I realised that this type isn't used at all now, so I've removed it! I have though done the check I was talking about in the real place where we're checking types after retrieving from the database

src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.ts
gdsandrewmccafferty: Yep, I had already done it, so I've removed the TODO

src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.ts
```diff
@@ -69,6 +69,40 @@ const getZendeskCustomSpaceSeparatedStringAsArray = (
...
+    ticketDetails,
+    getEnvAsNumber('ZENDESK_FIELD_ID_DATES')
+  )
+  if (ticketLegacyDateField) {
+    const returnValue = matchArrayParams(
+      [ticketLegacyDateField as string],
+      requestParams.dates
+    )
+    if (!returnValue) {
+      console.log(
```
SamBerry-GDS, line 90: This will get logged at the end of `ticketAndRequestDetailsDiffer` - if you think this is still useful to log could you use `logger` instead of `console.log`

src/sharedServices/s3/generateS3ObjectPrefixesForDateList.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      'firehose/2022/09/01/23',
+      'firehose/2022/09/02/00'
+    ]
+
+    const result = generateS3ObjectPrefixesForDateList([
+      '2022-08-21',
+      '2022-09-01'
+    ])
+    console.log('actual', result)
+    console.log('expected', expectedResult)
```
SamBerry-GDS, line 162: Probably don't need this as the test output should give us the same info

src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.ts
```diff
@@ -69,6 +69,40 @@ const getZendeskCustomSpaceSeparatedStringAsArray = (
...
+    ticketDetails,
+    getEnvAsNumber('ZENDESK_FIELD_ID_DATE_FROM')
+  )
+  const ticketDates = getZendeskCustomSpaceSeparatedStringAsArray(
+    ticketDetails,
+    getEnvAsNumber('ZENDESK_FIELD_ID_DATES')
+  )
+  if (ticketLegacyDateField) {
+    const returnValue = matchArrayParams(
+      [ticketLegacyDateField as string],
```
SamBerry-GDS, line 86: it might be more readable to do `as string | null` when we create ticketLegacyDateField so it's clear we know it's that because it's a single date rather than casting here.

src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.ts
gdsandrewmccafferty: You've cleverly spotted my debug code that wasn't supposed to be there... Removed now

src/sharedServices/s3/generateS3ObjectPrefixesForDateList.spec.ts
gdsandrewmccafferty: Again, should have taken it before, have done so now!

src/lambdas/initiateDataRequest/zendeskTicketDiffersFromRequest.ts
gdsandrewmccafferty: I like it. Done!

andylangridge (approved):

PR #236, opened 2023-02-13
Add test case for two dates to e2e tests

gdsandrewmccafferty:
This PR adds a test case to the e2e tests for a request for data from multiple dates.
It also changes the existing tests so they always pass data into the new date list field in Zendesk, because this will be what everything uses once we get this change into prod.


andylangridge (commented):
Did you mean to delete the `tests/integration-tests/constants/requests.ts` file?

It is in use by the `createAndApproveTicket` function (although I don't think that is used! So maybe delete that too?

gdsandrewmccafferty:
> Did you mean to delete the `tests/integration-tests/constants/requests.ts` file?
> 
> It is in use by the `createAndApproveTicket` function (although I don't think that is used! So maybe delete that too?

Yep, I did, and you're right, that other one isn't used at all, so I've deleted it as well.

andylangridge (approved):

PR #251, opened 2023-02-15
Feature/tt2 293 decryption lambda

gds-topickering:
Add new lambda to decrypt objects in the audit bucket and put decrypted objects in the analysis bucket.

Will be triggered by an s3 batch operation - this triggers the lambda for each object in the batch operation.


src/lambdas/decryptAndCopy/decryptS3Object.ts
```diff
@@ -0,0 +1,17 @@
+import { KmsKeyringNode } from '@aws-crypto/kms-keyring-node'
+import { buildDecrypt } from '@aws-crypto/decrypt-node'
+import { Readable } from 'stream'
+import { getEnv } from '../../utils/helpers'
+
+export const decryptS3Object = async (data: Readable): Promise<Buffer> => {
+  const keyring = new KmsKeyringNode({
+    generatorKeyId: getEnv('GENERATOR_KEY_ID')
+  })
+  // 'context' object may be required if added to corresponding encryption function
```
gdsandrewmccafferty, line 10: Not sure if we need this comment?

gds-topickering: I will remove it

gdsandrewmccafferty (approved):

gdsandrewmccafferty (approved):

PR #254, opened 2023-02-16
Update typescript and test dependencies

andylangridge:
* Update to latest version of Typescript
* Updated Yarn
* Updated Jest dependencies and jest config files as these broke after upgrading TS
* Renamed unit test files to `.test.ts` and kept integration/e2e tests as `.spec.ts`


tests/e2e-tests/jest.e2e.config.ts
```diff
@@ -24,7 +24,7 @@ const config: Config.InitialOptions = {
     ]
   ],
   setupFiles: ['<rootDir>/../shared-test-code/setup/setup.ts'],
-  testPathIgnorePatterns: ['/src/', '/tests/integration-tests/'],
+  testMatch: ['<rootDir>/*.spec.ts'],
```
gdsandrewmccafferty, line 27: Have you checked that just the e2e tests get run now when using this test config? It looks to me like this might just run everything with spec.ts in its name, which is obviously not what you want!

andylangridge: FIxed! I did some git jiggery pokery and was missing a change. Updated to `<rootDir>/**/*.spec.ts`

`<roorDir>` is the directory where the config file sits so all good on that front!

zaydgds (approved):

PR #261, opened 2023-02-20
Feature/tt2 294 decrypt files during data request

gdsandrewmccafferty:
Introduce a feature flag which lets us tell the processDataRequest lambda that it needs to decrypt files instead of just copying then, and that it should use the encrypted bucket as its data source.

The integration tests have been written to work off this feature flag and set up the data accordingly.


tests/integration-tests/test-suites/fullFlowDataQuery.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+      3,
+      '2022-05-01'
+    )
+    sendWebhookRequest(webhookRequest)
+    const downloadUrl = await pollNotifyMockForDownloadUrl(
+      webhookRequest.zendeskId
+    )
+    expect(downloadUrl.startsWith('https')).toBe(true)
+    const csvRows = await downloadResultsFileAndParseData(downloadUrl)
+    console.log('Got data ', csvRows)
```
andylangridge, line 26: Was this for debugging?

andylangridge (approved):

tests/integration-tests/test-suites/fullFlowDataQuery.spec.ts
gdsandrewmccafferty: It's in a test, I think it's probably useful to see it coming out at the end (and it'll only appear once so it's not clogging the logs up).

PR #262, opened 2023-02-20
Swap to use s3 logging bucket from infra stack

andylangridge:
Existing bucket can be safely removed as it is currently empty. It was missing a policy to allow the logging service to write any logs.


gdsandrewmccafferty:
⚠️ Is this logs bucket already in production? Fine otherwise, but a warning about merging if it isn't!

zaydgds (approved):

zaydgds (approved):

andylangridge:
Yep, bucket is in prod!

PR #267, opened 2023-02-22
TT2-335 removing allure

Ryan-Andrews99:
Removed allure integration from repo.


README.md
```diff
@@ -127,27 +127,7 @@ yarn test:e2e:dev
...
-```
-http://localhost:5252
-```
-
-An emailable version of the report is generated at the following location:
-
-```
-tests/reports/allure-reports/emailable-report-allure-docker-service.html
-```
+Running the tests creates a results file in JUnit format at `tests/reports/results`. Allure integration has been removed.
```
gdsandrewmccafferty, line 130: Minor point, I don't think we need to mention it's removed, let's just pretend it never existed 😄

gdsandrewmccafferty (approved):

README.md
Ryan-Andrews99: README updated

gdsandrewmccafferty (approved):

PR #270, opened 2023-02-27
Feature/tt2 337 decrypt unhappy paths

gdsandrewmccafferty:
This PR changes how we deal with the next step after triggering off the batch job that moves/decrypts files to the analysis bucket. 

We've changed over to an event-based setup where the batch job completion event triggers a new lambda called DataReadyForQuery (`src/lambdas/dataReadyForQuery/handler.ts`). Because this lambda can be triggered by any batch job, we check there that the tags we added to the job when it was created (in `src/sharedServices/bulkJobs/startTransferToAnalysisBucket.ts`) are present.

Also in this PR is the addition of a report for S3 batch jobs, so we can check for errors on individual files if they occur.


src/lambdas/dataReadyForQuery/jobWasSuccessful.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    )
+    expect(result).toEqual(false)
+  })
+
+  it.each`
+    totalNumberOfTasks | numberOfTasksFailed | numberOfTasksSucceeded | expected
+    ${10}              | ${0}                | ${10}                  | ${true}
+    ${10}              | ${1}                | ${9}                   | ${false}
+    ${0}               | ${0}                | ${0}                   | ${false}
+  `(
```
edgiles-gds, line 43: What a very nicely formatted table <3

edgiles-gds (approved):
Struggling to find anything off-kilter here; all looks good to my eyes. Nice refactoring and nice to see lots of cruft removal 😁

src/lambdas/dataReadyForQuery/jobWasSuccessful.test.ts
gdsandrewmccafferty: For that you can thank @Ryan-Andrews99 and some prettier rules...

PR #282, opened 2023-03-07
Tt2 407 removing unnecessary logging

gds-topickering:
In this PR, we've gone through the logging for the whole end to end journey and made it clearer, by adding a log line after every call to an external service, so we know where we are in the journey.

Where possible, we're also logging out things like message Ids for SQS messages, or e.g. Athena Query execution IDs.


zaydgds (approved):
seems sensible. Another pair of 👀

zaydgds (dismissed):
there is a function `sendSqsMessageWithStringBody`, could we add a 

```javascript
logger.info(`sensing message to queue ${queueUrl}`)
```

only caveat is that there is an ARN in the queue URL e.g. `https://sqs.eu-west-2.amazonaws.com/123456789/PublishToAccountsSQSDLQueue-dev` so we could maybe strip everything except the queue name

gdsandrewmccafferty:
> there is a function `sendSqsMessageWithStringBody`, could we add a
> 
> ```js
> logger.info(`sensing message to queue ${queueUrl}`)
> ```
> 
> only caveat is that there is an ARN in the queue URL e.g. `https://sqs.eu-west-2.amazonaws.com/123456789/PublishToAccountsSQSDLQueue-dev` so we could maybe strip everything except the queue name

Hmm, not sure that would add that much value, especially when we're already logging out when we've sent a message for a specific purpose (and we would then know what the queue URL is because it's in config?). Happy to discuss though if I've misunderstood!

src/lambdas/initiateAthenaQuery/handler.ts
```diff
@@ -72,14 +75,6 @@ const confirmQuerySqlGeneration = async (
     await updateZendeskTicketById(zendeskId, querySql.error, 'closed')
     throw new Error(querySql.error)
   }
-
-  if (querySql.sql) {
```
andylangridge, line 76: Would it not be useful to know what SQL was generated? This would help debugging queries, and nothing sensitive in the query itself!

src/sharedServices/bulkJobs/startGlacierRestore.ts
```diff
@@ -27,17 +27,13 @@ export const startGlacierRestore = async (
...
+  logger.info(`Starting Glacier restore`)
   const jobId = await createBulkGlacierRestoreJob(
     manifestFileName,
     manifestFileEtag,
     zendeskTicketId
   )
-  logger.info(
-    `Started Glacier restore for zendesk ticket with id '${zendeskTicketId}', with jobId '${jobId}'`
-  )
+  logger.info(`Started Glacier restore with jobId '${jobId}'`)
```
andylangridge, line 36: Why are we removing the ticket id here? Thought that was good for traceability

src/sharedServices/dynamoDB/dynamoDBUpdate.ts
```diff
@@ -42,8 +41,4 @@ export const updateQueryByZendeskId = async (
       `Failed to update item in db for zendesk ticket: ${zendeskId}`
     )
   }
-
-  logger.info(`Updated item in db: ${JSON.stringify(responseObject)}`)
```
andylangridge, line 46: Probably still useful to know a DB update was successful? Don't need the response object though.

src/sharedServices/s3/checkS3BucketData.ts
```diff
@@ -10,10 +10,7 @@ import { getAuditDataSourceBucketName } from './getAuditDataSourceBucketName'
 export const checkS3BucketData = async (
   dataRequestParams: DataRequestParams
 ): Promise<S3BucketDataLocationResult> => {
-  logger.info(
-    'Looking for S3 data using params',
-    JSON.stringify(dataRequestParams)
-  )
+  logger.info('Looking for S3 data')
```
andylangridge, line 13: This log probably a bit useless now without the data it's looking for.

src/sharedServices/s3/readS3DataToString.ts
```diff
@@ -2,17 +2,11 @@ import { GetObjectCommand, S3Client } from '@aws-sdk/client-s3'
 import consumers from 'stream/consumers'
 import { Readable } from 'stream'
 import { getEnv } from '../../utils/helpers'
-import { logger } from '../logger'
 
 export const readS3DataToString = async (
   bucketName: string,
   fileKey: string
 ): Promise<string> => {
-  logger.info(`reading file from bucket`, {
```
andylangridge, line 11: I think this is still a useful log to have, the key and bucket name would help spot if something was going wrong

src/sharedServices/zendesk/getZendeskTicket.ts
```diff
@@ -29,11 +27,5 @@ export const getZendeskTicket = async (id: string): Promise<ZendeskTicket> => {
     throw Error(interpolateTemplate('throwNotZendeskTicket', zendeskCopy))
   }
 
-  const ticketInfo = data.ticket
```
andylangridge, line 32: I think we should log out the ticket id that's not found here.

scripts/createAndApproveZendeskTicket.ts
```diff
@@ -54,10 +53,10 @@ const createAndApproveTicket = async () => {
     'OVERRIDE_EVENT_IDS'
   ])
   const ticketData = generateTicketData()
-  logger.info(`creating ticket with data ${JSON.stringify(ticketData)}`)
+  console.log('creating ticket')
```
andylangridge, line 56: As this is a script we can keep the JSON stringify bit, which is probably quite useful

src/lambdas/closeZendeskTicket/handler.ts
```diff
@@ -12,9 +12,6 @@ import {
 
 export const handler = async (event: SQSEvent, context: Context) => {
   initialiseLogger(context)
-  logger.info('Handling close zendesk ticket SQS event', {
```
andylangridge, line 15: Can we log out the ticket id that is being closed instead of deleting?

src/sharedServices/zendesk/updateZendeskTicket.ts
```diff
@@ -55,11 +55,7 @@ export const updateZendeskTicketById = async (
     }
   }
   try {
-    const data = await makeHttpsRequest(options, postData)
-    logger.info(
-      interpolateTemplate('zendeskSuccessful', loggingCopy),
-      JSON.stringify(data)
-    )
+    await makeHttpsRequest(options, postData)
```
andylangridge, line 58: Probably still nice to know the update was successful. Maybe use ticket id instead of all data?

src/utils/helpers.ts
```diff
@@ -34,7 +34,7 @@ export const tryParseJSON = (jsonString: string) => {
   try {
     return JSON.parse(jsonString)
   } catch (error) {
-    logger.error('Error parsing JSON: ', error as Error)
+    logger.error('Error parsing JSON')
```
andylangridge, line 37: Why are we deleting the error here? Which will include the stack trace etc.

src/lambdas/closeZendeskTicket/handler.ts
gds-topickering: An issue with logging out the ticket id here is that this log ran before validation as carried out on the event, and therefore you can't be sure that the event has a ticket id at this stage.  The next block of code parses the event and  checks for the zendesk ticket ID - if it's not there an error is thrown, if it is there (and all other expected fields are too) then the zendesk ID is appended to the logger and so will appear in logs.

src/lambdas/initiateAthenaQuery/handler.ts
gds-topickering: We took this out on the basis that the query SQL can be seen in the Athena console, even if the query fails, and so it would be possible to find it without needing to log it out. It's a matter of convenience perhaps - do we want to log information that can be found elsewhere?

src/sharedServices/bulkJobs/startGlacierRestore.ts
gdsandrewmccafferty: Because we're now putting the zendeskId into all logs as a property. That was the actually the original reason we put PowerLogger into the system, because it allows context logging (i.e. a property that follows through multiple logs without having to repeat it)

src/sharedServices/s3/readS3DataToString.ts
gdsandrewmccafferty: Not if it's only being used for one thing, which is to read the email recipient list

src/sharedServices/dynamoDB/dynamoDBUpdate.ts
gds-topickering: We removed this on the basis that a failed update would be logged so it may not be necessary to log a successful update

src/lambdas/initiateAthenaQuery/handler.ts
gdsandrewmccafferty: I think the consensus following our meeting earlier is that we can look this up in Athena, as Tom said, so I'm resolving this.

src/sharedServices/dynamoDB/dynamoDBUpdate.ts
gdsandrewmccafferty: Dealt with now by the fact we log out after every successful SDK operation

src/sharedServices/s3/checkS3BucketData.ts
gdsandrewmccafferty: Removed now

scripts/createAndApproveZendeskTicket.ts
gdsandrewmccafferty: Done

src/utils/helpers.ts
gdsandrewmccafferty: Have added this comment now

src/sharedServices/zendesk/updateZendeskTicket.ts
gdsandrewmccafferty: As part of the work to log out success of all external calls, this has been done now. The ZendeskId is always in the context, so it's there by default in the new message we added.

src/lambdas/processDataRequest/initiateDataTransfer.ts
```diff
@@ -31,17 +31,15 @@ export const initiateDataTransfer = async (
...
-
+  logger.info('Added data request to query request database')
   if (!glacierRestoreRequired && !copyFromAuditToAnalysisBucketRequired) {
     logger.info(interpolateTemplate('dataAvailableQueuingQuery', loggingCopy))
     await sendInitiateAthenaQueryMessage(dataRequestParams.zendeskId)
     return
   }
 
   if (glacierRestoreRequired) {
-    logger.info(interpolateTemplate('foundGlacierLocations', loggingCopy))
```
andylangridge, line 44: In this case, I think knowing that the Glacier Restore is going to happen is useful. Seeing as if it starts, we won't know it completes for a loooong time.

andylangridge: Actually, I can see this is logged in the function itself!

andylangridge (approved):

.eslintrc.json
```diff
@@ -9,5 +9,12 @@
     "jest/globals": true,
     "node": true
   },
-  "plugins": ["jest"]
+  "rules": { "no-console": "error" },
+  "plugins": ["jest"],
+  "overrides": [
```
SamBerry-GDS, line 14: Could we have overrides immediately after rules, just so it's really clear they are linked.

gdsandrewmccafferty: Yep, done!

SamBerry-GDS (approved):

PR #298, opened 2023-04-05
implement decryption via manual pii audit data request

zaydgds:


src/lambdas/initiateAthenaQuery/handler.ts
```diff
@@ -22,6 +24,15 @@ export const handler = async (
 
   const zendeskId = retrieveZendeskIdFromEvent(event)
   appendZendeskIdToLogger(zendeskId)
+
+  if (zendeskId.startsWith('MR')) {
+    logger.warn('Manual query detected, no need to run athena query')
```
gdsandrewmccafferty, line 29: This should be an `logger.info`, not warn, because this is a normal scenario you want to support

src/lambdas/initiateAthenaQuery/handler.ts
```diff
@@ -22,6 +24,15 @@ export const handler = async (
...
+    await publishToSNS(
```
gdsandrewmccafferty, line 30: Could we create a really simple function to do the notification of data ready, something like `sendManualQueryDataReadyMessage`, put it in a separate file, and mock that out instead? That way, we keep the complications in the handler to an absolute minimum, and the tests end up being simpler too.

src/lambdas/initiateAthenaQuery/handler.test.ts
```diff
@@ -174,4 +190,26 @@ describe('initiate athena query handler', () => {
     )
     expect(mockUpdateQueryByZendeskId).not.toHaveBeenCalled()
   })
+
+  it('returns an empty object as it is a manual query', async () => {
+    const testZendeskId = testManualAthenaQueryEvent.Records[0].body
+    const emailSNSTopicARN =
+      'arn:aws:sns:eu-west-2:123456789012:email-to-slack-topic'
+    when(getEnv)
```
gdsandrewmccafferty, line 198: Our pattern for mocking environment variables is that, rather than mocking out the `getEnv` utility, we instead add a line to `src/utils/tests/setup/testEnvVars.ts`.

template.yaml
```diff
@@ -1026,6 +1043,44 @@ Resources:
...
+    Type: AWS::KMS::Alias
+    Properties:
+      AliasName: !Sub 'alias/${AWS::StackName}/${Environment}/SNSKMSKey'
+      TargetKeyId: !Ref SNSKMSKey
+
+  EmailToSlackSNSTopic:
+    Type: AWS::SNS::Topic
+    Properties:
+      TopicName: !Sub '${AWS::StackName}-email-to-slack-topic'
+      KmsMasterKeyId: !GetAtt SNSKMSKey.Arn
```
gdsandrewmccafferty, line 1082: As above, I think you can just use the key referenced by the SSM parameter `CoreSnsKmsKeyArn`

zaydgds: re-worked the notification resources to live in the infra repo

PR #304, opened 2023-04-12
trigger decryption via manual pii audit data request

zaydgds:


src/sharedServices/sns/publishToSNS.ts
```diff
@@ -0,0 +1,13 @@
+import { SNSClient, PublishCommand } from '@aws-sdk/client-sns'
```
gdsandrewmccafferty, line 1: Can we add some tests for this? Apart from anything else, it'll push our test coverage down if we don't, and also every other similar bit of code that interacts with the AWS SDK currently has tests.

src/lambdas/initiateAthenaQuery/handler.test.ts
```diff
@@ -174,4 +185,18 @@ describe('initiate athena query handler', () => {
...
+    mockPublishToSNS.mockResolvedValue('messageID')
+    await handler(testManualAthenaQueryEvent, mockLambdaContext)
+    expect(mockPublishToSNS).toHaveBeenCalledWith(
+      'arn:aws:sns:eu-west-2:123456789012:email-to-slack-topic',
+      `Retrieved data for zendeskID: ${testZendeskId}`
+    )
+    expect(logger.info).toHaveBeenCalledWith(
+      'Manual query detected, no need to run athena query'
+    )
+    expect(confirmAthenaTable).not.toHaveBeenCalled()
```
gdsandrewmccafferty, line 200: If you're leaving the other dependencies in place, I think I'd like to see some assertions that they're also not called. At a minimum, you want to make sure `startQueryExecution` isn't called.

src/sharedServices/sns/publishToSNS.ts
zaydgds: added a test

src/lambdas/initiateAthenaQuery/handler.test.ts
zaydgds: added a wrapper

src/lambdas/initiateAthenaQuery/handler.test.ts
```diff
@@ -1,21 +1,6 @@
...
-import { getDatabaseEntryByZendeskId } from '../../sharedServices/dynamoDB/dynamoDBGet'
-import { createQuerySql } from './createQuerySql'
-import { updateQueryByZendeskId } from '../../sharedServices/dynamoDB/dynamoDBUpdate'
-import { ConfirmAthenaTableResult } from '../../types/athena/confirmAthenaTableResult'
-import {
-  dataPathsTestDataRequest,
-  noIdTestDataRequest,
-  testDataRequest
-} from '../../utils/tests/testDataRequest'
+import * as initiateQueryImportWrapper from './initiateQuery'
```
gdsandrewmccafferty, line 2: You don't need this any more, or the spy

src/lambdas/initiateAthenaQuery/handler.test.ts
```diff
@@ -42,152 +27,8 @@ jest.mock('../../sharedServices/sns/publishToSNS', () => ({
...
-    expect(mockUpdateZendeskTicket).toHaveBeenCalledWith(
-      testZendeskId,
-      'test athena error',
-      'closed'
-    )
-    expect(mockUpdateQueryByZendeskId).not.toHaveBeenCalled()
-  })
-})
-
 describe('tests related to running manual queries', () => {
```
gdsandrewmccafferty, line 32: You need a test for the "normal" scenario, where now you'll be asserting that the parameters from the handler are passed through correctly. You also need to reinstate some of the tests that are to do with the event body being malformed.

src/lambdas/initiateAthenaQuery/handler.ts
```diff
@@ -22,95 +16,14 @@ export const handler = async (
 
   const zendeskId = retrieveZendeskIdFromEvent(event)
```
gdsandrewmccafferty, line 17: I would put retrieveZendeskIdFromEvent back in it the handler, and then put the tests for the event structure being malformed here. It doesn't make sense to do that in the other file, and then export it.

src/lambdas/initiateAthenaQuery/initiateQuery.test.ts
```diff
@@ -0,0 +1,211 @@
...
+  Promise<ConfirmAthenaTableResult>
+>
+const mockUpdateZendeskTicket = updateZendeskTicketById as jest.Mock
+const mockGetDatabaseEntryByZendeskId = getDatabaseEntryByZendeskId as jest.Mock
+const mockCreateQuerySql = createQuerySql as jest.Mock
+const mockUpdateQueryByZendeskId = updateQueryByZendeskId as jest.Mock
+const mockStartQueryExecution = startQueryExecution as jest.Mock
+const mockPublishToSNS = publishToSNS as jest.Mock
+
+describe('initiate athena query handler', () => {
```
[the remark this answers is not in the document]
zaydgds: done

src/lambdas/initiateAthenaQuery/handler.test.ts
zaydgds: done

src/lambdas/initiateAthenaQuery/handler.test.ts
zaydgds: done

src/lambdas/initiateAthenaQuery/handler.ts
zaydgds: done

zaydgds: done

PR #315, opened 2023-05-04
Feature/tt2 469 swap out inline policies

gdsandrewmccafferty:
This is part of some remediation work we're doing following a pen test. To allow easier audit of permissions, we're removing an inline IAM policies and moving them out to their own resources


template.yaml
```diff
@@ -510,102 +576,124 @@ Resources:
...
-              Effect: Allow
-              Action:
-                - kms:GenerateDataKey
-                - kms:Encrypt
-                - kms:Decrypt
-              Resource: '{{resolve:ssm:AthenaQueryOutputBucketKmsKeyArn}}'
-            - Sid: Events
-              Effect: Allow
-              Action:
-                - events:PutEvents
```
gdsandrewmccafferty, line 586: This permission `PutEvents` was here originally, but when I moved it out of the inline policy, I got a Checkov warning for being too permissive (because of the `*`). I couldn't see what it was for, and all the integration tests pass without it being there, so I don't think we need it. Possibly something that was added in an early development phase that somebody thought was fixing something, but it actually wasn't.

andylangridge (approved):

PR #320, opened 2023-05-12
Feature/merge athena output bucket policies

gdsandrewmccafferty:
In the last set of commits we added a bucket policy to the TiCF integration to allow the test role to access the Athena Output Bucket. This had the unwanted side-effect of wiping out the permissions defined in the infra repo on this bucket which were there to allow query results to copy from it. This PR puts those permissions here instead.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -1318,29 +1318,49 @@ Resources:
               Bool:
                 'aws:SecureTransport': 'true'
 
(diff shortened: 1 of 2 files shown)


gdsandrewmccafferty:
Closing this because we're going to create a test lambda to read the data instead

PR #322, opened 2023-05-22
Feature/tt2 626 process data lambda

gds-topickering:
Amends calls to s3 to gain file prefixes when processing request and confirming location and storage class of target files to be copied to analysis bucket.
- no longer grabs files based on hour, but instead by day, in order to reduce number of calls to s3 and mitigate risk of lambda timing out.


src/sharedServices/s3/generateS3ObjectPrefixesForDateList.ts
```diff
@@ -1,48 +1,22 @@
...
+const validateDate = (dateParts: string[]) => {
+  const epochDate = Date.UTC(
+    parseInt(dateParts[0]),
+    parseInt(dateParts[1]) - 1,
+    parseInt(dateParts[2])
   )
-}
 
-const dedupeArray = (input: string[]) => [...new Set(input)]
+  if (isNaN(epochDate))
```
gdsandrewmccafferty, line 20: Don't think I see a test for this scenario? I would do that, since you're handling it specifically, as it's always good to keep the coverage up.

gds-topickering: I have now added tests to cover the input date being a non-date string and a nonsense date string

gdsandrewmccafferty (approved):

src/sharedServices/s3/generateS3ObjectPrefixesForDateList.ts
gdsandrewmccafferty: Looks good!

PR #323, opened 2023-05-26
Configure Lambda functions to run with VPC

andylangridge:
### Added VPC
- Added VPC config for Lambda functions
- Removed some Checkov skips
- Make SAM validate run linting in pipeline

### Refactored InitiateAthenaQuery Lambda
- Removed the call to the Glue service to check if the table exists - now handle the error when it occurs initiating the Athena query
- Updated the `StartQueryExecutionResult` type with the entire `Error` instead of just the error message
- Had to update the unit tests to account for the fact we no longer check the table exists


src/lambdas/initiateAthenaQuery/initiateQuery.test.ts
```diff
@@ -53,31 +39,24 @@ describe('initiateQuery', () => {
...
       sqlGenerated: true,
       sql: 'test sql string',
       queryParameters: ['123']
     })
-    mockUpdateQueryByZendeskId.mockResolvedValue('test db return object')
-    mockStartQueryExecution.mockResolvedValue({
+    mockStartQueryExecution.mockReturnValue({
       queryExecuted: true,
-      queryExecutionId: 'test_id'
+      queryExecutionId: '123'
```
andylangridge, line 52: Think I'm going to add this as a test constant, along with the query parameters string

gdsandrewmccafferty (approved):

gds-topickering (approved):
LGTM

PR #348, opened 2023-06-30
TT2-738-add-security-txt-redirect-to-sal

Ryan-Andrews99:
### This PR adds: 

- An Open API spec which includes the original `/zendesk-webhook` endpoint 
- Four new endpoints :
    - `/security.txt`,  `/.well-known/security.txt`, which redirect to https://vdp.cabinetoffice.gov.uk/.well-known/security.txt
    - `/thanks.txt` and `/.well-known/thanks.txt`, which redirect to https://vdp.cabinetoffice.gov.uk/.well-known/thanks.txt

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   "packageManager": "yarn@3.4.1",
   "private": true,
   "scripts": {
-    "build": "mkdir -p dist && cp -f open-api-specification.yaml dist/open-api-specification.yaml && tsc --noEmit && ts-node ./esbuild.config.ts",
+    "build": "yarn bundleApiSpecs && tsc --noEmit && ts-node ./esbuild.config.ts",
+    "bundleApiSpecs": "mkdir -p dist && cp -f open-api-specification.yaml dist/open-api-specification.yaml",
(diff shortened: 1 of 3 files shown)


zaydgds (dismissed):
im assuming you've deployed a feature branch and all is working well

package.json
```diff
@@ -10,7 +10,7 @@
   "packageManager": "yarn@3.4.1",
   "private": true,
   "scripts": {
-    "build": "tsc --noEmit && ts-node ./esbuild.config.ts",
+    "build": "mkdir -p dist && cp -f open-api-specification.yaml dist/open-api-specification.yaml && tsc --noEmit && ts-node ./esbuild.config.ts",
```
andylangridge, line 13: V minor thing, but you can chain yaml scripts if you take bits out into their own scripts.

So we could have `yarn bundleApiSpecs`, which did `mkdir -p dist && cp -f open-api-specification.yaml dist/open-api-specification.yaml`

Then this line could be:
```suggestion
    "build": "yarn bundleApiSpecs && tsc --noEmit && ts-node ./esbuild.config.ts",
```

andylangridge: Just makes these things easier to read and keeps them from getting too long.

Ryan-Andrews99: Cool didn't know that was a thing! I've added it in

andylangridge (approved):

PR #377, opened 2023-08-03
update script to use correct source path

zaydgds:
tweaking the file path as i was getting s3 copy errors


src/sharedServices/s3/copyS3Object.ts
```diff
@@ -6,16 +6,17 @@ import {
 import { getEnv } from '../../utils/helpers'
 
 export const copyS3Object = async (
-  fileName: string,
-  fileLocationPath: string,
-  fileDestinationBucket: string
+  Key: string,
```
gdsandrewmccafferty, line 9: Can you stick to the normal Camel case for function parameters here?

zaydgds: done

PR #388, opened 2023-09-04
Add script to copy data for manual data request

andylangridge:
* Adds a script to initiate the data retrieval process for a manual audit data request
* Since there was also a sendResult script, I moved this code into a shared folder `scripts/manualAuditDataRequests` along with some code for the new script (separated into sub-folders)


scripts/manualAuditDataRequestInitiateCopyAndDecrypt.ts
```diff
@@ -0,0 +1,36 @@
...
+    Math.floor(Math.random() * 1000000).toString()
+  )
+
+program.parse(process.argv)
+
+const options = program.opts()
+const dates: string[] = options.dates
+const zendeskId: string = options.zendeskId
+
+if ((!dates && !isStringArray(dates)) || !zendeskId) {
```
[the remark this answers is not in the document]
andylangridge: Good point, shall look at adding!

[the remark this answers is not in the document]
andylangridge: I updated the `isStringArray` function to `isDatesArray` and added some tests for it. Let me know what you think :)

scripts/manualAuditDataRequests/initiateCopyAndDecrypt/isDatesArray.test.ts
```diff
@@ -0,0 +1,48 @@
+import { isDatesArray } from './isDatesArray'
+
+describe('isDatesArray', () => {
+  it('should return true if the array is an array of dates', () => {
+    const testArray = ['2020-01-01', '2020-01-02']
```
gdsandrewmccafferty, line 5: Minor point, but you could inline all these variables, there's no need to declare something you're only using once, unless it adds clarity, which in this case I don't think it does.

andylangridge: Done

scripts/sharedServices/aws/sqs/sendSqsMessage.ts
```diff
@@ -0,0 +1,35 @@
+import {
```
gdsandrewmccafferty, line 1: As discussed, let's not copy and paste something that's nearly identical here, and just set an environment variable in code to let us use the version of the same thing in `src`

PR #389, opened 2023-09-05
Update send results script to not require environment variables

andylangridge:
The variables in this script were referencing the Analysis bucket instead of the Output bucket. Have updated the code so it no longer required environment variables, we just pass in an additional command line argument "environment", which is the only thing that changes within the bucket name between environments.

Updated the name of the zendesk ticket id argument, so it is a bit shorter, and matches the other script.


scripts/manualAuditDataRequests/sendResults/sendSQSMessageToCompletedQueue.ts
```diff
@@ -1,10 +1,10 @@
 import { sendSqsMessage } from '../../../src/sharedServices/queue/sendSqsMessage'
 import { sendManualQueryPayload } from './sendManualQueryPayload'
-import { getEnv } from '../../../src/utils/helpers'
 
 export const sendSQSMessageToCompletedQueue = async (
+  environment: string,
   manualQueryCompleteSqsPayload: sendManualQueryPayload
```
gdsandrewmccafferty, line 6: While we're here, could we make the `sendManualQueryPayload` match the standard uppercase for types rule?

andylangridge: Fixed! Move it into the types folder too

Ryan-Andrews99 (approved):

PR #434, opened 2024-04-26
TT2-1546 - increase concurrent executions for decrypt to speed up audit retrieval and prevent timeout issues

zaydgds:


template.yaml
```diff
@@ -566,6 +566,7 @@ Resources:
       CodeUri: dist/decryptAndCopy/
       Handler: handler.handler
       MemorySize: 1536
+      ReservedConcurrentExecutions: 50
```
eearcher-gds, line 569: where does the value of 50 come from?

zaydgds: completely random to be honest. i think i could probably make it way higher to be honest
so each month generate ~3000 files, i do bulk decryptions by quarters so 9000 files. account lambda concurrency 3000
10% of 3000 = 300
10% of 9000 = 900
i could probably make the concurrent executions 300 while leaving enough account concurrency for firehose. main goal of the concurrency controls in audit specifically is to ensure enough concurrency for firehose

eearcher-gds (approved):

dannymadden-gds (approved):

